PR #132, opened 2026-08-07
chore(deps): migrate @rtorcato/js-tooling@1 to @rtorcato/repo-tooling@3

rtorcato:
`@rtorcato/js-tooling` was renamed to `@rtorcato/repo-tooling` and the old
name has been **unpublished from npm**. This repo was the only remaining
consumer, pinned at `^1.0.9`, so a clean install could no longer resolve its
tooling at all.

## What changed

All six consumed subpaths, which all still exist under the new name:

| File | Subpath |
|---|---|
| `biome.json`, `apps/docs/biome.json` | `/biome` |
| `tsconfig.json` | `/typescript/base` |
| `vitest.config.mjs` | `/vitest/config` |
| `commitlint.config.mjs` | `/commitlint/config` |
| `build.mjs` | `/esbuild` |
| `release.config.mjs` | `/semantic-release/github` |

Plus the `.github/dependabot.yml` group pattern and the README / Docusaurus
sibling-project links.

Historical `CHANGELOG.md` entries and the issue titles in `TODO.md` keep the
old name — they are a record of what happened, not references to resolve.

## Two majors, one real consequence

`2.0.0` moved 39 packages from `dependencies` to optional `peerDependencies`.
This repo already declares every one it actually uses as a direct
devDependency, so nothing went missing — but it is why `pnpm-lock.yaml`
shrinks by ~1200 lines. `3.0.0` was the rename itself.

## One thing this surfaced

`@biomejs/biome` is bumped `^2.3.0` → `^2.5.0`. The shared preset uses
`linter.rules.preset`, which is a Biome **2.5** key, so the previously
resolved 2.4.9 rejected the config outright with `Found an unknown key
\`preset\``. The `$schema` URLs in both biome.json files move to 2.5.7 to
match.

That is an upstream bug, not a local one: repo-tooling advertises
`"@biomejs/biome": "^2.0.0"` as its peer range, which is wider than its own
preset supports. Any consumer on 2.0–2.4 hits the same wall. Filed separately
against repo-tooling.

## Not a breaking change

Deliberately `chore(deps):` with no `!` and no `BREAKING CHANGE:` footer.
Everything here is devDependencies and build config — the published API and
the shipped bundle are byte-for-byte unaffected, so this should not cut a
major.

## Verification

`biome check` (139 files), `tsc --noEmit`, **371 tests across 50 files**,
`build-prod`, and a direct `import()` of `release.config.mjs` and
`commitlint.config.mjs` to confirm the two presets that no other step
exercises actually resolve.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,7 +25,7 @@ updates:
       - "*prettier*"
       - "vitest*"
       - "typescript"
-      - "@rtorcato/js-tooling"
+      - "@rtorcato/repo-tooling"
       - "biome"
       - "husky"
       - "lint-staged"
--- a/README.md
+++ b/README.md
@@ -237,7 +237,7 @@ Contributions are welcome! Please read [CONTRIBUTORS.md](./CONTRIBUTORS.md) for 
 ## Related packages
 
 - [@rtorcato/browser-common](https://github.com/rtorcato/browser-common) — Browser Web API wrappers (clipboard, observers, storage, etc.)
-- [@rtorcato/js-tooling](https://github.com/rtorcato/js-tooling) — Project scaffolding for TypeScript libraries (Biome, Vitest, Husky, semantic-release)
+- [@rtorcato/repo-tooling](https://github.com/rtorcato/repo-tooling) — Project scaffolding for TypeScript libraries (Biome, Vitest, Husky, semantic-release)
 
 ## License
 
--- a/apps/docs/biome.json
+++ b/apps/docs/biome.json
@@ -1,7 +1,7 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.4.9/schema.json",
+  "$schema": "https://biomejs.dev/schemas/2.5.7/schema.json",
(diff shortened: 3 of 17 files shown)


rtorcato:
Pushed a follow-up: CI caught a peer dependency my local verification missed.

`pnpm coverage` failed with `Cannot find dependency '@vitest/coverage-v8'`.
The shared vitest preset hardcodes `coverage.provider: 'v8'`, and under
js-tooling@1 that package arrived transitively as a real dependency. The
2.0.0 peer-dependency change makes it optional, so it is now absent unless
declared — which corrects the claim in the PR description that this repo
already declared every peer it uses. It declared all of them but this one.

The `@vitest/coverage-istanbul` that *was* in devDependencies turned out to
be dead weight: nothing overrides the preset's provider, so coverage has been
running v8 the whole time and the istanbul package was never loaded. Swapped
rather than added, so only the provider actually in use is installed.

`pnpm coverage` now exits 0 with the configured thresholds met.

PR #149, opened 2026-08-11
fix(types): return the wrapper type from debounce/throttle and type to()'s error slot as unknown

### LINKED ISSUE #144: fix(types): debounce/throttle return T unsoundly; to() types its error slot as any
Two inference bugs in the same vein, both found while adding type tests for #76. The type tests there encode current behaviour deliberately, with comments — flip them when this lands.

## 1. `debounce` / `throttle` are unsound (the serious one)

`src/functions/index.ts:26` and `:41`:

```ts
export function debounce<T extends (...args: any[]) => void>(fn: T, wait: number): T {
	let timeout: ReturnType<typeof setTimeout> | undefined
	return function (this: any, ...args: any[]) {
		clearTimeout(timeout)
		timeout = setTimeout(() => fn.apply(this, args), wait)
	} as T          // <- the wrapper returns undefined, but is cast to T
}
```

The constraint is `=> void`, but TypeScript's return-type-void rule lets a **value-returning** function satisfy it. So `T` infers as the full signature including its return type, the result is typed as returning that type, and the `as T` cast papers over the fact that the wrapper returns `undefined`.

### Repro — compiles clean, produces NaN

```ts
const d = debounce((n: number) => n * 2, 10)
const result = d(21)              // typed as number; actually undefined
const doubled: number = result * 2
console.log(doubled)
```

```
$ tsc --noEmit --strict probe.ts
$ echo $?
0
$ tsx probe.ts
NaN
```

Arithmetic on `undefined` with zero type errors. Any call site that uses a debounced function's return value is silently wrong.

### Fix

Return the wrapper's real type rather than `T`:

```ts
export function debounce<T extends (...args: any[]) => void>(
	fn: T,
	wait: number
): (...args: Parameters<T>) => void
```

This is a **breaking change** for anyone currently relying on the (incorrect) return type — but what it breaks is code that was already broken at runtime. Worth calling out in the release notes.

## 2. `to()` types the error slot as `any`

`src/promises/index.ts:16`:

```ts
export async function to<T>(promise: Promise<T>): Promise<[any, T | undefined]>
```

The point of the `[error, value]` tuple is to make error handling explicit, and `any` defeats that at the type level — `err.message`, `err.foo.bar`, anything goes.

`unknown` is the correct shape and forces a narrow at the call site. Also breaking, for callers doing `err.message` directly.

## Suggested sequencing

Both are breaking type-level changes with no runtime component, so they pair naturally in one major. Neither changes emitted JS.

## Notes

- `debounce`'s JSDoc says `@returns {Function}`, so the intent was always "returns a wrapper", not "returns T".
- Found via #76. Related stale references in that issue: `promises/retry` does not exist in `src/`, and `groupBy` lives in `arrays`, not `objects`.

rtorcato:
`debounce` and `throttle` were declared to return `T`. TypeScript's
return-type-void rule lets a value-returning function satisfy the
`(...args: any[]) => void` constraint, so `T` inferred the full signature
including its return type and the `as T` cast hid the fact that the
wrapper returns `undefined`. `debounce((n: number) => n * 2, 10)(21) * 2`
typechecked and produced `NaN`. Both now return
`(...args: Parameters<T>) => void`, which is what the wrapper actually is,
and the casts are gone.

`to()` typed its error slot as `any`, which defeats the point of the
`[error, value]` tuple — `err.foo.bar` was accepted. It is now `unknown`,
forcing a narrow at the call site.

Neither change alters emitted JavaScript.

BREAKING CHANGE: `debounce` and `throttle` no longer claim to return the
wrapped function's return type. Code reading a debounced or throttled
call's return value now fails to compile — that code was already getting
`undefined` at runtime. `to()`'s error slot is `unknown` instead of `any`,
so call sites doing `err.message` directly must narrow first
(`err instanceof Error ? err.message : String(err)`).

Closes #144

### DIFF
--- a/src/functions/functions.test-d.ts
+++ b/src/functions/functions.test-d.ts
@@ -15,7 +15,6 @@ describe('functions — types', () => {
 	it('debounce preserves the wrapped function parameters', () => {
 		const fn = (_a: string, _b: number) => {}
 		const debounced = debounce(fn, 100)
-		expectTypeOf(debounced).toEqualTypeOf<typeof fn>()
 		expectTypeOf(debounced).parameters.toEqualTypeOf<[string, number]>()
 		expectTypeOf(debounced).returns.toEqualTypeOf<void>()
 		// @ts-expect-error — arguments are still checked
@@ -25,19 +24,17 @@ describe('functions — types', () => {
 	it('throttle preserves the wrapped function parameters', () => {
 		const fn = (_event: { x: number }) => {}
 		const throttled = throttle(fn, 100)
-		expectTypeOf(throttled).toEqualTypeOf<typeof fn>()
 		expectTypeOf(throttled).parameters.toEqualTypeOf<[{ x: number }]>()
 		expectTypeOf(throttled).returns.toEqualTypeOf<void>()
 		// @ts-expect-error — arguments are still checked
 		throttled({ y: 1 })
 	})
 
-	it('debounce/throttle keep a non-void return type even though the wrapper drops it', () => {
-		// Current (unsound) behaviour: the `=> void` constraint accepts value-returning
-		// functions, so `T` — and therefore the returned type — still says `number`,
-		// while at runtime the wrapper returns undefined. Asserted as-is, not endorsed.
+	it('debounce/throttle return void even when the wrapped function returns a value', () => {
+		// The wrapper never forwards `fn`'s return value, so the type must not claim it does.
 		const fn = (n: number) => n * 2
-		expectTypeOf(debounce(fn, 10)).returns.toEqualTypeOf<number>()
-		expectTypeOf(throttle(fn, 10)).returns.toEqualTypeOf<number>()
+		expectTypeOf(debounce(fn, 10)).returns.toEqualTypeOf<void>()
+		expectTypeOf(throttle(fn, 10)).returns.toEqualTypeOf<void>()
+		expectTypeOf(debounce(fn, 10)).parameters.toEqualTypeOf<[number]>()
 	})
 })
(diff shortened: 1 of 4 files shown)


rtorcato (commented):

Reviewed for injection, secrets, unsafe shell/SQL, and supply-chain risk.

This is a type-only change (`debounce`/`throttle` return type, `to()` error slot `any` → `unknown`) — no emitted JS changes, no new dependencies, no runtime logic touched. No shell execution, no string interpolation into commands/queries, no secrets in the diff. The `unknown` narrowing for `to()`'s error slot is a security-positive change (forces callers to validate error shape instead of trusting `any`).

No findings. Clean.

rtorcato (commented):

Diff matches the fix proposed in #144 exactly: `debounce`/`throttle` now return `(...args: Parameters<T>) => void` instead of the unsound `T` cast (`as T` removed in both), and `to()`'s error slot is `unknown` instead of `any`. Neither changes emitted JS, consistent with the PR description.

Type tests are updated coherently: the old `toEqualTypeOf<typeof fn>()` assertions are correctly dropped (no longer true since the return type now legitimately differs from `fn`'s), replaced with explicit `.returns.toEqualTypeOf<void>()` and `.parameters.toEqualTypeOf<...>()` checks. The `to()` test-d.ts assertion is updated to `unknown` to match.

BREAKING CHANGE footer is present and accurately describes the compile-time-only breakage, which is required for semantic-release given this is a `fix(types)` commit with a breaking change.

No correctness issues found. Nit: JSDoc `@returns {Function}` on `debounce`/`throttle` could be tightened to reflect the new explicit signature, but that's cosmetic and not worth blocking on.

PR #161, opened 2026-08-11
docs: consume @rtorcato/shared-docs for the sibling-family list

### LINKED ISSUE #103: docs: consume @rtorcato/shared-docs for the sibling-family list
Migrate this repo's docs site to the shared sibling-family list in
[@rtorcato/shared-docs](https://github.com/rtorcato/shared-docs), removing the
locally duplicated family list. Reference implementation: rtorcato/cf-common#87.

## Steps
1. Add the git dependency (not on npm; `dist/` is committed, so no install-time build step):
   `pnpm --filter ./apps/docs add -D "github:rtorcato/shared-docs"`
2. `docusaurus.config.ts`: replace the local `PROJECT_FAMILY` array with
   `import { projectFamilyItems } from '@rtorcato/shared-docs'` and
   `const PROJECT_FAMILY = projectFamilyItems()`.
3. `src/pages/index.tsx`: replace the local `SIBLINGS` array with
   `import { siblings } from '@rtorcato/shared-docs'` and
   `const SIBLINGS = siblings('@rtorcato/<this-package>')`.
4. Delete any local `family.ts` / inline family data.
5. `pnpm --filter ./apps/docs build` — the rendered nav dropdown, footer, and
   sibling grid should be unchanged.

## Notes
- If this repo's family data (taglines/accents) is newer than what's in
  shared-docs, update `src/family.ts` in shared-docs first, then rebuild +
  commit its `dist/` and push.
- Shared components (InstallTabs + the mobile-sidebar swizzles) will be
  centralized in a later phase — out of scope for this issue.

rtorcato:
Migrates the docs site to the shared sibling-family list in [@rtorcato/shared-docs](https://github.com/rtorcato/shared-docs), removing the locally duplicated family data.

## Changes

- Added `@rtorcato/shared-docs` as a dev dependency of `apps/docs` (git dependency — not on npm; its `dist/` is committed, so there is no install-time build step).
- `apps/docs/docusaurus.config.ts`: the local `PROJECT_FAMILY` array is now `projectFamilyItems()`, and `GITHUB_PROFILE` is imported from the package instead of being declared locally. Both navbar-dropdown and footer consumers are unchanged.
- `apps/docs/src/pages/index.tsx`: the local `Sibling` type and `SIBLINGS` array are now `siblings('@rtorcato/js-common')`.
- No local `family.ts` existed; those two arrays were the only inline family data.

## Rendered output differs — expected

The shared list is ahead of the local copies, so the nav dropdown, footer column and sibling grid now also include `supabase-common` and `db-x`, and pick up the `js-tooling` → `repo-tooling` rename. This is the point of the migration, not a regression.

## Verification

- `pnpm --filter ./apps/docs build` succeeds. Inspecting the built `index.html`: the nav dropdown and footer render all family entries plus "All on GitHub →", and the sibling grid renders 9 cards with `@rtorcato/js-common` correctly excluded.
- `pnpm typecheck` and `pnpm check` pass (also enforced by the pre-commit hook).
- `pnpm --filter ./apps/docs typecheck` fails on a pre-existing `TS5101 baseUrl is deprecated` error from `@docusaurus/tsconfig`; confirmed identical on `main` without these changes.

Shared components (InstallTabs + the mobile-sidebar swizzles) are a later phase and out of scope here.

Closes #103

### DIFF
--- a/apps/docs/docusaurus.config.ts
+++ b/apps/docs/docusaurus.config.ts
@@ -1,21 +1,13 @@
 import type * as Preset from '@docusaurus/preset-classic'
 import type { Config } from '@docusaurus/types'
+import { GITHUB_PROFILE, projectFamilyItems } from '@rtorcato/shared-docs'
 import { themes as prismThemes } from 'prism-react-renderer'
 
-// The @rtorcato open-source family. Surfaced as a navbar "Projects" dropdown
-// (Docusaurus renders navbar items in the mobile menu too) and in the footer,
-// so every sibling site cross-links to the rest. Keep in sync across repos.
-const GITHUB_PROFILE = 'https://github.com/rtorcato'
-const PROJECT_FAMILY = [
-	{ label: 'js-common', href: 'https://rtorcato.github.io/js-common/' },
-	{ label: 'api-common', href: 'https://rtorcato.github.io/api-common/' },
-	{ label: 'browser-common', href: 'https://rtorcato.github.io/browser-common/' },
-	{ label: 'db-common', href: 'https://rtorcato.github.io/db-common/' },
-	{ label: 'cf-common', href: 'https://rtorcato.github.io/cf-common/' },
-	{ label: 'react-common', href: 'https://github.com/rtorcato/react-common' },
-	{ label: 'swift-common', href: 'https://rtorcato.github.io/swift-common/' },
-	{ label: 'repo-tooling', href: 'https://rtorcato.github.io/repo-tooling/' },
-]
+// The @rtorcato open-source family, from the single source of truth in
+// @rtorcato/shared-docs. Surfaced as a navbar "Projects" dropdown (Docusaurus
+// renders navbar items in the mobile menu too) and in the footer, so every
+// sibling site cross-links to the rest.
+const PROJECT_FAMILY = projectFamilyItems()
 
 const config: Config = {
 	title: 'js-common',
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -29,6 +29,7 @@
 		"@docusaurus/tsconfig": "^3.8.1",
 		"@docusaurus/types": "^3.8.1",
 		"@playwright/test": "^1.60.0",
+		"@rtorcato/shared-docs": "github:rtorcato/shared-docs",
 		"@types/react": "^19.0.0",
 		"docusaurus-plugin-typedoc": "^1.4.2",
 		"typedoc": "^0.28.19",
--- a/apps/docs/src/pages/index.tsx
+++ b/apps/docs/src/pages/index.tsx
@@ -1,5 +1,6 @@
 import Link from '@docusaurus/Link'
 import useBaseUrl from '@docusaurus/useBaseUrl'
+import { siblings } from '@rtorcato/shared-docs'
 import Layout from '@theme/Layout'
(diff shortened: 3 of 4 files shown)


rtorcato (commented):

Verified both specific risk points from the task:

1. **`GITHUB_PROFILE`** — now imported from `@rtorcato/shared-docs` instead of being a local const. The diff only touches the declaration/import lines in `docusaurus.config.ts`; the two consumer sites (navbar dropdown, footer) are untouched, so they still resolve against the same identifier. No dangling reference to the old local const remains in the diff.

2. **Sibling exclusion** — `siblings('@rtorcato/js-common')` in `index.tsx` passes this repo's own package name, so js-common correctly excludes itself from its own grid. Matches the pattern in issue #103's instructions (`siblings('@rtorcato/<this-package>')`).

Other observations:
- Diff is a clean 1:1 swap of local arrays for the shared package calls, matches the reference pattern from issue #103 and the linked `cf-common#87` precedent.
- `package.json` / `pnpm-lock.yaml` correctly add `@rtorcato/shared-docs` as a git-hosted devDependency with no install-time build step, consistent with the issue's notes.
- The PR body explains a rendered-output diff (adds `supabase-common`, `db-x`, `js-tooling`→`repo-tooling` rename) as intentional (shared-docs is ahead of the old local copies) — reasonable given the migration's purpose, not a regression.
- Disclosed pre-existing `TS5101 baseUrl is deprecated` failure on `apps/docs typecheck` — out of scope per instructions, not blocking.
- Nit: `pnpm-lock.yaml` includes an unrelated `minimatch 10.2.5 → 10.2.6` bump in the typescript-eslint snapshot, likely incidental lockfile churn from regenerating the lockfile — harmless.

No defects found. Clean.

rtorcato (commented):

**Verdict: no defects — 0 critical/high/medium, 1 info.**

Reviewed as a supply-chain change: adds `@rtorcato/shared-docs` as a `github:` dependency instead of a registry package.

- **Pinning**: `pnpm-lock.yaml` resolves the spec `github:rtorcato/shared-docs` to a fixed commit tarball — `https://codeload.github.com/rtorcato/shared-docs/tar.gz/5fb0ae8ff63da6dcc027b2931ee7df3f7a48ce60` — with a recorded `integrity: sha512-...` hash. It is not floating on a branch ref; pnpm will refuse to install if the tarball content ever changes under that commit SHA. This is the correct pattern for a git dependency.
- **Blast radius**: added under `devDependencies` in `apps/docs/package.json`. devDependencies are never included in the dependency graph of a package's consumers regardless of whether the workspace is published or private, so this reaches no downstream consumer of `@rtorcato/js-common` — it's confined to the docs build. (The diff doesn't touch `private`, so I can't confirm that field directly, but it's moot given the dependency type.)
- **CI gate**: the `dependencies` check (lockfile supply-chain policy) passes (43s, green) on this PR.
- No secrets, no shell/SQL construction, no injection surface in the diff — it's config/import changes plus generated content in `docusaurus.config.ts` and `index.tsx`.
- Info-only: the lockfile also carries an incidental `minimatch 10.2.5 → 10.2.6` bump from the resolver, unrelated to this change and not a concern.

This is a normal, acceptable pattern: dev-only git dependency on the author's own repo, pinned to a commit SHA with integrity, in a devDependency-only path.

Issue #174, opened 2026-08-11
Settle the module boundaries — no more renames after this stage


rtorcato:
The package ships 46 subpath modules. Once the API is called stable, moving a helper between modules is a breaking change for every consumer that imported it — so the boundaries have to be settled *during* beta, while breaking changes are still allowed.

Known friction already documented in the README: `tryCatch` means different things in `/errors` (renamed to `tryWithFallback` in 2.0) and `/try` (the Result-pattern helper). That kind of same-name-different-module overlap is exactly what this issue is for.

- Audit for helpers that could plausibly live in two modules (`formatting` vs `strings`, `date` vs `datetime` vs `time`, `numbers` vs `math`, `validation` vs `boolean`).
- Decide the home for each, and rename or move *now* rather than after 1.0.
- Record the decisions so the next contributor does not re-litigate them.

Gate for the Beta milestone: after this closes, module paths are frozen.

rtorcato:
🤖 *Automated — triage via ai-issue-loop. Posted under the owner's account by an agent; not a human message. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

Dropping `ai-ready` — this one is not agent work in its current shape.

**Why an agent cannot finish it.** The deliverable is a *decision*, not an implementation: "decide the home for each" helper across 46 subpath modules, where every choice is a breaking change for consumers and is explicitly frozen after this closes. An agent can tell you that `formatting` and `strings` overlap; it cannot tell you which name you want to live with for the life of 1.0. A PR that picked those answers on its own judgement would need re-litigating field by field in review, which costs more than deciding them up front.

**What would make it automatable.** Split it in two:

1. *Audit only* — an agent enumerates every exported symbol across the 46 modules, flags name collisions (the `tryCatch` case), near-duplicates, and helpers whose module is arguable, and posts the table with a recommendation per row. No code changes. This is ordinary agent work and would make the decision cheap.
2. *Apply the agreed moves* — once you have signed off on the table, the renames, re-exports, `package.json` `exports` updates, and README/doc changes are mechanical, and an agent can do that in one PR.

Label step 1 `ai-ready` and the loop will pick it straight up.

**Not terminal** — the issue itself is worth keeping open as the Beta gate; it just needs step 1 carved off before an agent can help.

PR #182, opened 2026-08-11
docs(modules): add a hand-written guide above the generated reference

### LINKED ISSUE #176: A docs-site guide per module, not just generated API reference
All 46 modules have a page under `apps/docs/docs/modules/`, but they are generated by `scripts/generate-module-docs.mjs` — a signature list, not a guide. A stable API is supposed to be a *documented* one.

- For each module, add hand-written prose above the generated reference: what the module is for, when to reach for it over a neighbouring module, and a worked example.
- Keep the generated reference below it, so `pnpm docs:generate` stays non-destructive (verify it does not clobber hand-written sections — if it does, that is part of this issue).
- Prioritise the modules with overlap problems (see #174).

Gate for v1.0: "documented public API" is one of the three things this milestone means.

rtorcato:
Closes #176

Every module page under `apps/docs/docs/modules/` now opens with prose — what the module is for, when to reach for it over a neighbouring module, and a worked example — followed by the generated API reference and a "See also" list. All 45 module pages have all three; `overview.md` is untouched. The overlap-prone modules (#174) call the collision out explicitly: date/time/datetime, logger/logging, try/errors.

## Regeneration is non-destructive

`scripts/generate-module-docs.mjs` no longer skips existing files. It fences its output between `<!-- generated:exports -->` markers and rewrites only that block, so the hand-written guide around it survives. A page with no markers is left fully alone (`date`, `numbers`, `strings`, `uuid`). A second `pnpm docs:generate` is a no-op:

```
generate-module-docs: created 0, updated 0, skipped 4 (no marker block)
```

`spliceGeneratedBlock` is covered by `scripts/generate-module-docs.test.mjs`.

## Extraction bugs fixed

The generated block is now authoritative, which exposed three bugs that had previously been papered over by hand-editing the tables:

- The JSDoc regex could span `*/`, so a file-header comment was harvested as the summary for the first exported symbol (`mime-types` `extensions`, `FileExtension`).
- `firstSentence` ended the sentence at `e.g.` (`mime-types` `MimeType`).
- Tag stripping ran inside code spans, eating `Success<T>` and `<br>`.

JSDoc summaries in `colors`, `datetime` and `time` were corrected or marked up to match. Those are comment-only changes — no runtime behaviour changes anywhere in this PR.

## Verification

`pnpm verify` clean: `tsc --noEmit`, `biome check` (153 files), 387 tests passed / 3 skipped. All relative links between module pages resolve.

Not verified locally: the Docusaurus build. `apps/docs` has no `node_modules` in the agent worktree, so `pnpm --filter docs build` cannot run here — the `docs.yml` workflow is the check. `docusaurus.config.ts` sets `markdown.format: 'detect'`, so `.md` pages parse as CommonMark and the HTML-comment markers are safe there.


rtorcato (commented):

**Verdict: clean — no criticals/highs/mediums.**

Scope: `scripts/generate-module-docs.mjs` (+ new `generate-module-docs.test.mjs`), 45 generated `apps/docs/docs/modules/*.md` pages, and 3 JSDoc comment-only edits in `src/colors`, `src/datetime`, `src/time`. No `package.json` / lockfile changes — no dependency or supply-chain surface here.

- **Injection**: none. All inputs to the generator (`package.json#exports` keys, `src/<name>/index.ts` JSDoc) are repo-owned, trusted, build-time data — not attacker-controlled at runtime. No `child_process`, no `eval`, no SQL.
- **Path handling**: `docPath = resolve(docsDir, \`${sub}.md\`)` — `sub` is derived from the package's own `exports` map, not external input, so there's no traversal vector to worry about.
- **Regex changes** (`JSDOC_BLOCK`, `firstSentence`, `escapeForMarkdownTable`): reviewed for ReDoS. The new `JSDOC_BLOCK` alternation `(?:[^*]|\*(?!\/))*` and the `firstSentence` lookbehinds are all fixed/linear — no nested-quantifier overlap, so no catastrophic-backtracking risk. `escapeForMarkdownTable` now only strips tags outside backtick spans and still escapes `|`, which is correct for the markdown-table output it targets.
- **Non-destructive regeneration** (`spliceGeneratedBlock`): correctly returns `null` when marker comments are absent, so hand-written pages are never silently clobbered — this is the property the linked issue (#176) asked to be verified, and it's covered by a real test (`generate-module-docs.test.mjs`).
- **Secrets**: none found; example snippets reference `process.env.WEBHOOK_SECRET` / `DATABASE_URL` only as illustrative placeholders, nothing hardcoded.
- CLAUDE.md's "register new modules in `package.json#exports`" rule doesn't apply — no new modules were added, only doc/comment content.

No defects to block on.

rtorcato (commented):

Reviewed `gh pr diff 182` against issue #176 and the repo `CLAUDE.md`. Verified the recovered/uncommitted-work concern specifically: traced `src/colors/index.ts` and `scripts/generate-module-docs.mjs` end to end.

**Correctness**
- `spliceGeneratedBlock`/`buildBlock`/`generate()` in `scripts/generate-module-docs.mjs` are coherent: marker-fenced splice, "no markers → skip" fail-safe, and the new `created/updated/skipped` counters all match the PR's stated behaviour. Confirmed idempotency and the marker-fenced splice logic against the new `scripts/generate-module-docs.test.mjs` (both cases pass: replace-only-marked-block, and null-on-no-markers).
- Independently re-ran the two regex fixes (non-greedy `JSDOC_BLOCK` that stops at the first real `*/`, and the `e.g./i.e./etc` negative-lookbehind in `firstSentence`) against representative strings — both behave as described, including correctly *not* truncating at the abbreviation while still ending at the real sentence period.
- `src/colors/index.ts`'s `isValidHex` JSDoc fix matches the corrected `colors.md` row exactly. The `matchingTextColor` row also changed in `colors.md` with no corresponding source diff in this PR — checked the full current source and confirmed that JSDoc was already correct on `main` before this PR; the `.md` page was just stale until now. Coherent, not a leftover/half-applied edit.
- All 45 module pages sampled show the `<!-- generated:exports -->`/`<!-- /generated:exports -->` fence correctly wrapping only the Import+Exports block, hand-written prose above it, "See also" below — consistent with the issue's non-destructive-regeneration requirement.

**Nit**
- The PR description attributes the "`e.g.` truncates the sentence" bug specifically to `mime-types`' `MimeType`/`extensions`/`FileExtension`, but the diff shows no change to those doc rows or to `src/mime-types/index.ts` — the current generated rows for those symbols were already correct pre-PR. Harmless (the regex fix is real and independently verified above), just an imprecise attribution in the description.

No correctness, security, or convention issues found. Tabs/`export function`-style/biome conventions followed throughout the script changes.

rtorcato:
🤖 *Automated — conflict resolution by an agent, posted under the owner's account. Not a human message.*

Rebased onto current `main`. **This PR was not abandoned** — it stopped exactly where the pipeline is designed to stop (both reviewer labels, waiting on a human merge), and then #184 landed underneath it and turned it red.

`mergeable` is back to `MERGEABLE`, and all five required checks pass: `lint`, `typecheck`, `build`, `test (22)`, `test (24)`.

## Conflicts and how they were resolved

- **`formatting.md` / `math.md` deleted.** #184 removed both modules, so the hand-written guides this branch added for them documented nothing.
- **`datetime.md`, `html.md`** conflicted *inside* the `generated:exports` block. Rather than hand-merging the tables I re-ran `pnpm docs:generate`, since this branch's own generator is the source of truth for that region. Its improved `escapeForMarkdownTable` also fixed the mangled `textToHtml` summary for free.
- **`datetime.md`'s hand-written example** imported `secondsBetween` from `datetime`; #184 moved it to `time`. Now imports from both.
- **`events.md`'s hand-written example** used `once`, which #184 renamed `onceEvent`, plus a line on why the two names were separated.

I also merged the remote branch head, which this worktree was missing — no force-push.

## Verified

`typecheck`, `biome`, 364 tests, `build-prod`, and a full Docusaurus build — that last one being the real check that no page still links the two deleted guides.

I wrote a checker over every `import { … } from '@rtorcato/js-common/…'` in the module pages. Only `events.once` was breakage from #184, and it is fixed. It also surfaced pre-existing phantom names (`strings.removeEmojis`, `uuid.generateUUID`, `uuid.isValidUUID`) — **left alone deliberately**, they are tracked in #185 and are not this PR's job.

## Still-red checks — all pre-existing, none required

- **`commitlint`** — a 73-character line, one over the cap. Fix is in #188, which disables `footer-max-line-length` the same way `body-max-line-length` already was. Merge that and this goes green.
- **`Test Coverage`** — was already failing here at 84.07% before #184; now 83.46%. Tracked in #189.
- **`CodeQL`** — 15 open high alerts, all on `main`, not introduced here. Tracked in #190.

Ready for a human merge.

rtorcato:
Unblocked — `mergeStateStatus` went `BLOCKED` → `UNSTABLE`.

The blocker was never the earlier conflict or the 15 pre-existing alerts on `main`. It was the `code-scanning-main` ruleset (`security_alerts_threshold: high_or_higher`) firing on **4 alerts this PR introduced** in `scripts/generate-module-docs.mjs` — the one file it rewrites. The rewrite shifted the line numbers enough that CodeQL scored them as new.

Both fixes are smaller or less wrong than the code they replace:

- `escapeForMarkdownTable` escapes `<` instead of stripping `/<[^>]*>/`. A one-pass strip leaves a tag behind on nested input (`<<b>>` → `<b>`) and silently eats the `<T>` from an unbackticked `Success<T>`.
- It escapes `\` in the same pass as `|`. Escaping only the pipe turned `a\|b` into `a\\|b` — an escaped backslash plus a live pipe, so the row broke anyway.
- `generate()` reads and treats `ENOENT` as absent instead of asking `existsSync` and then writing, dropping both `js/file-system-race` instances.

Regenerating produces byte-identical pages, so no documentation content moved. `escapeForMarkdownTable` is now exported and covered by two tests pinning the nested-tag and backslash cases.

The one remaining red check is **Test Coverage**, which is not a required check and is tracked separately as #189.

PR #194, opened 2026-08-12
chore(commitlint): drop the local body/footer overrides

### LINKED ISSUE #191: chore(commitlint): drop the local body/footer overrides now that repo-tooling disables them
_(AI-authored issue.)_

`@rtorcato/repo-tooling` 3.9.1 sets `body-max-line-length` and `footer-max-line-length` to `[0]` in the shared preset ([repo-tooling#401](https://github.com/rtorcato/repo-tooling/issues/401), [PR #402](https://github.com/rtorcato/repo-tooling/pull/402)) — which is exactly what this repo's `commitlint.config.mjs` override does locally. The override is now redundant and can be deleted; fixing it upstream instead of copying the override to a seventh sibling repo was the stated point.

## Steps, in this order

1. `pnpm update @rtorcato/repo-tooling` — the lockfile currently resolves **3.4.0**. The `^3.0.0` specifier in `devDependencies` already allows 3.9.1, so no `package.json` edit is needed.
2. Reduce `commitlint.config.mjs` to:

   ```js
   import config from '@rtorcato/repo-tooling/commitlint/config'
   export default config
   ```

Reversing that order re-enables the 72-char cap in the window between the two steps.

## Heads-up

The bump spans **3.4.0 → 3.9.1**, not a patch, so it carries every other preset change since 3.4.0 (biome, tsconfig, CI presets). Expect unrelated drift to surface in the same PR — split it out if it gets noisy.

## Verify

A commit with a >72-character `BREAKING CHANGE:` footer still passes `pnpm exec commitlint`.

rtorcato:
`@rtorcato/repo-tooling` 3.9.1 sets `body-max-line-length` and
`footer-max-line-length` to `[0]` in the shared preset, which is exactly
what this repo's `commitlint.config.mjs` was doing locally. The override
is now redundant.

## Changes

- Bump `@rtorcato/repo-tooling` 3.4.0 → 3.9.1 (lockfile).
- Reduce `commitlint.config.mjs` to a bare re-export of the preset.
- `pnpm update` also raised the devDependency floor from `^3.0.0` to
  `^3.9.1`. The issue said no `package.json` edit was needed, but this one
  is worth keeping: with the override deleted, any resolution below 3.9.1
  would silently re-enable the 72-character cap. The floor now encodes the
  requirement.

## Preset drift, 3.4.0 → 3.9.1

The bump spans five minors, so I diffed the three presets this repo
actually consumes:

- `tooling/typescript/*` — byte-identical.
- `tooling/biome/biome.json` — one change, `vcs.enabled` `false` → `true`
  (with `useIgnoreFile: true` already set). Biome now honours `.gitignore`.
- `tooling/commitlint/commitlint.mjs` — the two rules above, `[2, 'always', 72]` → `[0]`.

No other drift surfaced, so there is nothing to split out.

## Verification

Run against a real install of 3.9.1 in the worktree, not the shared
`node_modules`:

- `pnpm exec commitlint` accepts a commit whose `BREAKING CHANGE:` footer
  is well over 72 characters — the issue's stated check. Exit 0.
- `pnpm verify` (typecheck + biome check + vitest): clean, 363 tests
  passing, 150 files checked.

## Note for the reviewer

The pre-push hook initially failed in this agent worktree because
`node_modules` was a symlink into the main checkout, which made pnpm
resolve bin paths to a doubled path
(`js-common-worktrees/js-common-worktrees/...`). That is a worktree-setup
issue, unrelated to this change. Fixed by giving the worktree its own
install; the hook then ran and passed on the push above.

Closes #191

### DIFF
--- a/commitlint.config.mjs
+++ b/commitlint.config.mjs
@@ -1,19 +1,3 @@
 import config from '@rtorcato/repo-tooling/commitlint/config'
 
-// config-conventional caps body AND footer lines at 72 characters. That rule is
-// aimed at `git log` in an 80-column terminal; it buys nothing here and reliably
-// fails agent-written commit messages, leaving PRs permanently amber on a check
-// that is not required to merge. Amber checks nobody acts on are worse than none.
-// Subject-line rules stay enforced — those decide the release type.
-//
-// The footer half matters more than it looks: `BREAKING CHANGE:` is a footer, so
-// the one place a message most needs room to explain itself was the one place
-// still capped. PR #182 and #184 each burned a CI run on a 73-character line.
-export default {
-	...config,
-	rules: {
-		...config.rules,
-		'body-max-line-length': [0, 'always', Number.POSITIVE_INFINITY],
-		'footer-max-line-length': [0, 'always', Number.POSITIVE_INFINITY],
-	},
-}
+export default config
--- a/package.json
+++ b/package.json
@@ -243,7 +243,7 @@
     "@biomejs/biome": "^2.5.0",
     "@commitlint/cli": "^21.2.1",
     "@commitlint/config-conventional": "^21.2.0",
-    "@rtorcato/repo-tooling": "^3.0.0",
+    "@rtorcato/repo-tooling": "^3.9.1",
     "@semantic-release/changelog": "^7.0.0",
     "@semantic-release/commit-analyzer": "^13.0.1",
     "@semantic-release/git": "^11.0.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -40,8 +40,8 @@ importers:
         specifier: ^21.2.0
         version: 21.2.0
       '@rtorcato/repo-tooling':
-        specifier: ^3.0.0
-        version: 3.4.0(30d8862d0d43ceafc31e73019da20b9e)
+        specifier: ^3.9.1
+        version: 3.9.1(30d8862d0d43ceafc31e73019da20b9e)
       '@semantic-release/changelog':
         specifier: ^7.0.0
         version: 7.0.0(semantic-release@25.0.9(supports-color@8.1.1)(typescript@6.0.3))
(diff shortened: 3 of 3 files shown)


rtorcato (commented):

Diff matches the issue's ask: `commitlint.config.mjs` is reduced to a bare re-export of the preset, which is simpler than the issue's own suggested two-line version and functionally equivalent (preset's `[0]` disables the rule same as the deleted `[0, 'always', Infinity]` override). Lockfile bump 3.4.0 → 3.9.1 looks consistent (specifier, resolution, and snapshot entries all match; integrity hash changed as expected for the new version).

On the devDependency floor (`^3.0.0` → `^3.9.1`): the issue said this edit was unnecessary since `^3.0.0` already resolves 3.9.1. The PR raised it anyway, reasoning that once the local override is gone, any future resolution below 3.9.1 (e.g. via a fresh `pnpm update` without a pinned lockfile) would silently reintroduce the 72-char cap with no local signal. That's a legitimate correctness argument for the invariant this PR relies on, not scope creep — it's a one-line, low-risk constraint tightening, not a defect. Judgment call worth a note, not a blocker.

No issues found with the commitlint config change itself, and the PR's own diffing of the three consumed presets (typescript, biome, commitlint) between 3.4.0 and 3.9.1 to rule out unrelated drift is reasonable due diligence per the issue's "split it out if it gets noisy" heads-up — nothing noisy surfaced.

rtorcato (commented):

**Verdict: clean — no criticals, highs, mediums. 1 info.**

Scope: `@rtorcato/repo-tooling` devDependency bump 3.4.0 → 3.9.1, lockfile update, and a `commitlint.config.mjs` simplification to re-export the upstream preset now that it sets `body-max-line-length`/`footer-max-line-length` to `[0]` itself.

**Supply-chain check (the substance of this review):**
- The lockfile diff shows only `@rtorcato/repo-tooling`'s own version/resolution/integrity entries changing. Its declared dependencies (`chalk@5.6.2`, `commander@15.0.0`) are unchanged between 3.4.0 and 3.9.1 — no new transitive packages, no version drift in its dependency tree. Nothing snuck in beyond the package bump itself.
- devDependency only — no runtime/production code path is affected. Blast radius is limited to local tooling (commitlint, biome config, build) and CI.
- The `^3.0.0` → `^3.9.1` floor bump in `package.json` is a good catch by the author: without it, a fresh install could re-resolve to something below 3.9.1 and silently reintroduce the 72-char cap now that the local override is deleted. Correct reasoning, not scope creep.
- No secrets, no shell/SQL construction, no injection surface anywhere in this diff — it's config + lockfile only.

**Info:** The PR body itself is candid that this wasn't validated against the shared root `node_modules` (worktree symlink issue), but says verification was run against a real install of 3.9.1 in an isolated worktree with `pnpm exec commitlint` and `pnpm verify` passing. No action needed, just noting the verification method for the record.

No blocking issues.

Issue #193, opened 2026-08-12
Agent worktrees are missing apps/docs/node_modules, so the docs build (the only broken-link check) cannot run


rtorcato:
## Problem

`.claude/settings.json` symlinks only the root `node_modules` into an agent
worktree:

```json
"worktree": {
  "symlinkDirectories": ["node_modules"]
}
```

`apps/docs/node_modules` is not included, so in a fresh worktree:

```
$ pnpm --filter @rtorcato/js-common-docs build
docusaurus: command not found
```

## Why it matters more than it looks

The docs build is **the only check that catches broken documentation links** —
Docusaurus fails the build on them. An agent working on docs in a worktree
therefore cannot run the one verification that matters for its change, and the
failure mode is a missing binary rather than anything that reads as "you are
missing a link check", so it gets skipped rather than fixed.

Hit twice now: once while resolving #176/#182 and once again clearing the
CodeQL alerts on that same branch. Both times the workaround was manual:

```
ln -s /Volumes/RtorcatoSSD/WORK/js-common/apps/docs/node_modules apps/docs/node_modules
```

## Fix

Probably just add the nested path:

```json
"symlinkDirectories": ["node_modules", "apps/docs/node_modules"]
```

**This needs verifying, which is why it is an issue rather than a one-line PR.**
I could not confirm that `symlinkDirectories` accepts a nested path rather than
only a top-level directory name, and shipping an unverified entry risks
breaking worktree creation for every agent. Confirm against the Claude Code
settings schema, then check a freshly created worktree actually gets both
symlinks before merging.

If nested paths are not supported, the fallback is a `SessionStart` hook that
creates the symlink, or running `pnpm install --filter` in the worktree.

## Related

- #192 makes `.gitignore` actually ignore these symlinks — they are currently
  untracked, so `git add -A` in a worktree commits them.

PR #195, opened 2026-08-12
test: cover the ISO week, groupBy, detectLanguage and checkEnv gaps

### LINKED ISSUE #189: Test Coverage check is below threshold and has been red for a while
The `Test Coverage` workflow fails on every PR. It is not a required status check, so nothing blocks on it — which is why it has gone unnoticed.

```
ERROR: Coverage for lines (83.46%) does not meet global threshold (85%)
ERROR: Coverage for functions (94.87%) does not meet global threshold (95%)
ERROR: Coverage for statements (83.64%) does not meet global threshold (85%)
```

## It predates #184, but #184 made it worse

| | lines | statements | functions |
|---|---|---|---|
| #182 before the #184 merge | 84.07% ❌ | 84.29% ❌ | pass |
| #182 after the #184 merge | 83.46% ❌ | 83.64% ❌ | 94.87% ❌ |
| threshold | 85% | 85% | 95% |

So it was already under on two metrics. #184 deleted `src/formatting` and `src/math` — both small and fully covered — and removing 100%-covered code drags the average down, which pushed `functions` under too.

## Where the gaps are

From a local `pnpm coverage` run, worst first:

| Module | % Lines | Note |
|---|---|---|
| `logger` | 20% | pino instance; hard to test meaningfully |
| `datetime` | 31% | `getIsoWeek` / `getIsoWeekInfo` tests **exist but are commented out** in `src/datetime/index.test.ts` |
| `os` | 44% | environment probes |
| `process` | 55% | environment probes |
| `env` | 65% | |
| `i18n` | 66% | `detectLanguage` branches |
| `arrays` | 75% | `index.ts:168-172` |

Note the local total reads higher than CI's; whatever CI includes that a local run does not should be pinned down as part of this, because a threshold you cannot reproduce locally is a threshold nobody can act on.

## The judgement call — not for an agent to decide alone

Two defensible fixes, and they are not equivalent:

1. **Write the missing tests.** Best value is `datetime` — those tests are already written and merely commented out; find out why before simply uncommenting, since someone disabled them for a reason.
2. **Lower the threshold to match reality.** Legitimate for `logger`/`os`/`process`, which are thin wrappers over environment globals where a test asserts little beyond the mock.

Do (1) where the code is genuinely testable. Do **not** silently lower thresholds to green the check — if that is the answer for the probe modules, propose the number and the reasoning in the PR and let a human take it.

rtorcato:
The Test Coverage workflow has been red on every PR: lines and statements sat
under the 85% global threshold and #184 pushed functions under 95% too.

Raises real coverage rather than moving the thresholds:

- datetime: `getIsoWeek` / `getIsoWeekInfo` had tests, but three successive
  attempts were left commented out. Both functions read the *local* calendar
  fields of their input before re-projecting onto UTC, so the UTC-constructed
  dates every attempt used (`new Date('2025-01-01')`, `new Date(Date.UTC(...))`)
  shift a day in any zone behind UTC. Rewritten with local components, which
  makes them timezone-independent, plus the week-53 and ISO-year-boundary cases.
- arrays: `groupBy` was entirely untested.
- i18n: `detectLanguage` only had a test that accepted either answer. Now stubs
  `navigator` to pin the browser branch, the LANG branch, and both defaults.
- env: `checkEnv` was entirely untested, including the ZodError message and the
  non-Zod rethrow.

Totals go from 83.13% lines / 83.23% statements / 95.51% functions to
88.12% / 87.88% / 97.11%. No threshold was touched and no file was excluded.

Closes #189

### DIFF
--- a/src/arrays/index.test.ts
+++ b/src/arrays/index.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest'
-import { chunk, compact, first, flatten, last, shuffle, unique } from './index'
+import { chunk, compact, first, flatten, groupBy, last, shuffle, unique } from './index'
 
 describe('first', () => {
 	it('returns the first element', () => {
@@ -39,6 +39,27 @@ describe('chunk', () => {
 describe('compact', () => {
 	it('removes falsy values', () => {
 		expect(compact([0, 1, false, 2, '', 3, null, undefined])).toEqual([1, 2, 3])
+	})
+})
+
+describe('groupBy', () => {
+	it('groups by a computed string key', () => {
+		expect(groupBy([1, 2, 3, 4], (n) => (n % 2 === 0 ? 'even' : 'odd'))).toEqual({
+			odd: [1, 3],
+			even: [2, 4],
+		})
+	})
+
+	it('groups objects by a property', () => {
+		const items = [{ type: 'a' }, { type: 'b' }, { type: 'a' }]
+		expect(groupBy(items, (x) => x.type)).toEqual({
+			a: [items[0], items[2]],
+			b: [items[1]],
+		})
+	})
+
+	it('returns an empty object for an empty array', () => {
+		expect(groupBy([], (n: number) => n)).toEqual({})
 	})
 })
 
--- a/src/datetime/index.test.ts
+++ b/src/datetime/index.test.ts
@@ -1,6 +1,8 @@
 import { describe, expect, it } from 'vitest'
 import {
 	formatDateTimeLocal,
+	getIsoWeek,
+	getIsoWeekInfo,
 	getTimezoneOffset,
 	nowIso,
 	parseIsoDateTime,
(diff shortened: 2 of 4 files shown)


rtorcato:
🤖 *Automated — implementer via ai-issue-loop. Posted under the owner's account by an agent; not a human message. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

**On "the local total reads higher than CI's".** It does not, once you run the same command CI runs. `pnpm run coverage` at `98a85fc` reproduces CI exactly: 83.13% lines / 83.23% statements / 95.51% functions. The gap in the issue's numbers comes from `scripts/generate-module-docs.mjs` being inside the coverage include set — it sits at 26.82% statements and is the single worst entry in the report, so any mental model that only looks at `src/**` reads several points high. Nothing CI-only is involved; the threshold *is* reproducible locally.

**Nothing was excluded and no threshold was moved.** The four modules fixed here were all genuinely testable, so option (1) from the issue covered the whole gap on its own — the judgement call on `logger`/`os`/`process` never had to be made. Those three are still the worst in the report and are left exactly as they were:

| Module | % Lines |
|---|---|
| `scripts/generate-module-docs.mjs` | 27.63% |
| `logger` | 20% |
| `os` | 44.44% |
| `process` | 54.54% |
| `system` | 58.62% |

Two things a human may still want to decide, neither of which blocks this PR:

1. Whether `scripts/**` belongs in library coverage at all. Excluding it would be defensible (it is a docs generator, not shipped code) but it is a scope change, not a test fix, so I left it in — and the totals clear the thresholds with it included.
2. Whether `Test Coverage` should become a required status check now that it is green. That is what let it rot unnoticed in the first place.

**One environment note.** `pnpm run coverage` cannot run in an agent worktree as-is: `node_modules` is symlinked to the main checkout, and the `.bin` shims there resolve to a path one directory level shallower, so every shim misses. It is unrelated to this PR and probably worth its own issue — see also #193.

rtorcato (commented):

Verdict: clean, no security concerns.

- **Secrets**: `env` fixtures use obviously-fake values (`https://example.com`, `PORT: '8080'`) passed as plain objects directly into `checkEnv(schema, {...})` — never touching `process.env` or a real `.env` file. No hardcoded credentials or API keys.
- **Ambient environment**: The pre-existing `OLD_ENV` save/restore pattern in `env/index.test.ts` is untouched; the new `checkEnv` tests don't mutate `process.env` at all, so there's nothing to leak across files.
- **Global stubbing (i18n)**: `navigator` is stubbed with `vi.stubGlobal` and cleaned up in a scoped `afterEach` via `vi.unstubAllGlobals()`, and the original `LANG` value is restored (or deleted if previously unset) rather than left mutated. No cross-file leakage.
- **Injection / shell / SQL**: None — this PR only adds `describe`/`it`/`expect` blocks calling in-repo pure functions (`groupBy`, `getIsoWeek`, `getIsoWeekInfo`, `checkEnv`, `detectLanguage`). No `child_process`, `eval`, or string-built queries.
- **Dependencies / CI**: Diff touches only four `*.test.ts` files. No `package.json`, lockfile, or `.github/workflows/*` changes rode along; the `zod` import in the env test is consuming an existing dependency (schema/`checkEnv` already exist in `src/env/index.ts`), not adding a new one.

Nit (non-blocking): the `checkEnv` rethrow test builds a fake schema object cast via `as unknown as typeof schema` — harmless in a test file, just noting it's a type-safety shortcut rather than a real `ZodSchema`.

rtorcato (commented):

**Scope check:** Diff touches only four test files (`src/arrays/index.test.ts`, `src/datetime/index.test.ts`, `src/env/index.test.ts`, `src/i18n/i18n.test.ts`). No `vitest.config`/coverage config, no `include`/`exclude` sets, no `istanbul ignore` pragmas, no threshold anywhere in the diff. This is genuinely tests-only, as claimed — confirms the PR isn't moving the goalposts on #189.

**Timezone reasoning for `getIsoWeek`/`getIsoWeekInfo` (the main risk area):** the claim is that `new Date(year, month, day)` (local components) is what makes these tests TZ-independent, versus the previously-commented-out `new Date('2025-01-01')` / `Date.UTC(...)` forms. That holds: local-component construction and `getFullYear`/`getMonth`/`getDate` reads both operate in the runtime's local zone, so the same y/m/d round-trips regardless of the CI runner's TZ — whereas a UTC-parsed string read back via local getters can shift a day west of UTC, which is exactly the bug the old commented-out tests hit. I hand-checked the calendar arithmetic in the new assertions (2025-01-01 = week 1, 2025-05-26 = week 22, 2025-12-31 → ISO week 1 of 2026, 2020-12-31 = week 53, 2021-01-01 → week 53 of 2020) and all of them are correct ISO week/year values, not just numbers that happen to pass in one zone.

**Assertion quality:** all four files assert real outcomes, not just "ran without throwing":
- `groupBy` — checks actual grouped output for numeric and object keys, plus the empty-array case.
- `checkEnv` — checks parsed/coerced values, the missing-key error message text, that the stack is stripped, the non-Zod rethrow path, and `RootApiEnvSchema` defaults.
- `detectLanguage` — uses `vi.stubGlobal('navigator', ...)` to pin the browser branch (including region-subtag stripping), the `LANG` branch, and both default-fallback branches, replacing the old test that accepted either answer.

No defects found. Style (tabs, single quotes, no unneeded semicolons) matches the existing test files.

PR #196, opened 2026-08-12
chore(worktree): symlink apps/docs/node_modules into agent worktrees

### LINKED ISSUE #193: Agent worktrees are missing apps/docs/node_modules, so the docs build (the only broken-link check) cannot run

rtorcato:
Adds `apps/docs/node_modules` to `worktree.symlinkDirectories`, so an agent worktree can run the docs build — the only check that catches broken documentation links.

Closes #193

## The verification the issue asked for

The issue held this back as an issue rather than a one-line PR because nested paths in `symlinkDirectories` were unconfirmed. They work:

- Claude Code rejects only absolute paths and `..` segments. Every entry is joined onto both the repo root and the worktree root, so a nested relative path resolves correctly. Every failure branch warns and continues, so a bad entry cannot break worktree creation for other agents.
- Symlinking runs after `git worktree add` has checked out the tree, so the `apps/docs` parent directory already exists when the link is made.

Checked against a freshly created worktree, not just read from the source:

- both `node_modules` and `apps/docs/node_modules` are symlinked;
- `git check-ignore -v` traces both to `.gitignore:5`, so `git add -A` in a worktree stays clean (#192);
- `pnpm --filter @rtorcato/js-common-docs build` runs to completion (exit 0, `Generated static files in "build"`), where it previously died with `docusaurus: command not found`.

## One thing that came up, and why it is not this PR

Midway through, every pnpm binary broke in *all four* agent worktrees — `tsc` resolved to a nonexistent doubled path (`js-common-worktrees/js-common-worktrees/ai-191-.../node_modules/.pnpm/...`), surfacing as a confusing `MODULE_NOT_FOUND` in the pre-commit hook.

That is not caused by symlinking, and not by this change. The root `node_modules` symlink has been in place all along. The cause is running `pnpm install` **inside** a worktree whose `node_modules` is a symlink: pnpm writes through the link and re-points the *shared* root `.bin` shims at that worktree's virtual store. The main repo kept working only because the bad path happened to resolve back through another worktree's symlink — so the shared install had become dependent on a worktree that could be deleted at any time.

`pnpm install --frozen-lockfile` from the main repo restored it (node_modules only; no lockfile or tracked-file change), and typecheck/lint pass in the worktree again — this commit went through the pre-commit hook normally.

Worth noting that this PR reduces the pressure that causes that breakage: the missing docs binary is exactly what tempts an agent to run `pnpm install` inside a worktree. Making the binary resolve without an install is the point. A guardrail against in-worktree `pnpm install` is a separate concern and probably deserves its own issue.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -15,6 +15,6 @@
     ]
   },
   "worktree": {
-    "symlinkDirectories": ["node_modules"]
+    "symlinkDirectories": ["node_modules", "apps/docs/node_modules"]
   }
 }


rtorcato (commented):

**Verdict: clean.**

Diff is exactly the one-line change the issue describes — `.claude/settings.json` `worktree.symlinkDirectories` gains `"apps/docs/node_modules"` alongside the existing `"node_modules"` entry. Nothing else in the file changed.

Checked specifically for the privilege-relevant risks this file could carry:

- No `permissions` block present or touched — no allow/deny rule added, removed, or weakened.
- No `hooks` key added — no new command wired to run automatically.
- No other top-level key introduced. The diff is a single array literal edit, 1 addition / 1 deletion, one file.
- The new path is repo-relative (`apps/docs/node_modules`), has no leading `/` and no `..` segments, and resolves inside the repo tree (mirrors the existing `apps/docs` app directory) — it does not point outside the repo or at a sensitive location (no home dir, no `.ssh`, no `.env`, no `.git`).

No injection, no secrets, no shell/SQL construction, no dependency or supply-chain changes — this PR only edits harness config, no code paths touched.

No defects found.

rtorcato (commented):

Diff is a single-line, syntactically valid addition to `symlinkDirectories` (`"node_modules"` → `"node_modules", "apps/docs/node_modules"`), matching exactly the fallback fix the issue proposed. No logic to break, no repo convention (exports registration, build scripts, commit style) that this touches.

The issue explicitly gated this on verification of four things, and the PR body addresses all four with specific, checkable detail rather than a bare "I tested it":
- nested-path support — explained via the mechanism (absolute/`..` paths rejected, others joined onto repo root and worktree root; failures warn-and-continue so a bad entry can't break other agents' worktrees)
- both symlinks present in a fresh worktree
- `git check-ignore -v` tracing both to `.gitignore:5` (a specific line reference, tying back to #192)
- `pnpm --filter @rtorcato/js-common-docs build` going from the exact `docusaurus: command not found` failure named in the issue to exit 0 with the actual Docusaurus success string (`Generated static files in "build"`)

That's not pasted CI/log output, so it can't be independently re-verified from the diff alone, but the specificity (exact error strings, a `.gitignore` line number, the concrete success message) reads as genuine verification rather than hand-waving — which is what the issue asked for.

The tangential pnpm/`.bin` shim corruption note is clearly scoped out as a separate, pre-existing issue unrelated to this change, and correctly not folded into this diff.

No defects found. Nit only: a raw terminal transcript would have been stronger evidence than a narrated summary, but not blocking.

PR #197, opened 2026-08-12
fix: make the ReDoS-flagged regexes linear-time

### LINKED ISSUE #190: CodeQL: 15 open high-severity alerts, incl. the designated sanitizer
`gh api repos/rtorcato/js-common/code-scanning/alerts?state=open` returns **15 open alerts, all rated high**, all on `refs/heads/main`. The `CodeQL` check shows red on PRs, but only reports the subset "new in code changed by this pull request", so a PR touching an untouched file shows green and the backlog stays invisible.

`CodeQL` is not a required status check, so none of this blocks a merge today. Note the `code-scanning-main` ruleset *is* active, which is worth confirming does not eventually gate pushes to `main`.

## The one that deserves a decision

`src/security/index.ts:19` — `sanitizeString`, which carries **four** alerts: `js/incomplete-multi-character-sanitization` ×2, `js/bad-tag-filter`, `js/polynomial-redos`.

```ts
return str.replace(/<script.*?>.*?<\/script>/gi, '').replace(/on\w+\s*=\s*(['"]).*?\1/gi, '')
```

CodeQL is right — a blocklist over `<script>` and `on*=` is trivially bypassable (`<scr<script>ipt>`, `onerror=x` unquoted, ``).

The JSDoc **already says so**, and that is deliberate:

> Blocklist-based, so it is bypassable — not a substitute for a real sanitizer such as DOMPurify when rendering untrusted HTML.

But `MODULE-BOUNDARIES.md` made `security` the owner of this name specifically because "sanitising is a security concern". So the module whose name implies safety ships a helper documented as unsafe. That tension is the actual issue, and it is a judgement call:

- keep it, and lean harder on naming/docs (`stripScriptTagsNaive`?), or
- drop it and point at DOMPurify, or
- accept it as-is and dismiss the alerts with a rationale.

**That decision is yours, not an agent's** — which is why this issue is not labelled `ai-ready`.

## The rest — mechanical

Nine `js/polynomial-redos` alerts on regexes that can see untrusted input:

- `src/emails/index.ts:15`
- `src/url/index.ts:97`, `98`
- `src/validation/index.ts:91`
- `src/currency/index.ts:185`
- `src/html/index.ts:84`

Plus `js/incomplete-multi-character-sanitization` on `src/html/index.ts:66` (`stripHtmlTags`, `/<[^>]*>/g`) and three alerts in `scripts/generate-module-docs.mjs` (build-time tooling over trusted input — lowest real risk of the set).

These are bounded-quantifier fixes and are ordinary work once the `sanitizeString` question above is settled, since the answer may change what `html`/`security` should look like.

## Suggested split

Settle `sanitizeString` first, then a follow-up issue can carry the ReDoS fixes as `ai-ready`.

rtorcato:
Closes #190

Fixes the mechanical ReDoS alerts, documents the one weakness that cannot be
fixed by rewriting a scan, and leaves the alerts that need an owner decision.

## Fixed: ReDoS (quadratic → linear)

All of these were ambiguous unbounded quantifiers — a long crafted string costs
quadratic time.

| Alert | Location | Fix |
|---|---|---|
| #8, #15 `js/polynomial-redos` | `emails.isValidEmail`, `validation.isEmail` | domain labels now exclude `.`, so each split is unique |
| #13, #14 `js/polynomial-redos` | `url.joinUrl` | index scan instead of `/^\/+\|\/+$/g` |
| #16 `js/polynomial-redos` | `currency.parsePrice` | `\d` instead of `\d+` — identical test, no unbounded quantifier |
| #26 `js/polynomial-redos` | `html.stripHtmlTags` | one `indexOf` scan for `<` then `>` |
| #10 `js/polynomial-redos` | `html.containsHtml` | same scan |

**The email regexes get stricter.** `foo@bar..com` and `foo@bar.com.` were
accepted before and are rejected now (the old `[^\s@]+` swallowed dots). That is
the behaviour change that makes this a `fix:` rather than a `chore:`. Both cases
are pinned by new tests.

## Not fixed, now documented: `stripHtmlTags` incomplete sanitization (#27)

**Correcting an earlier version of this PR description, which claimed this alert
was fixed. It was not.**

The `indexOf` rewrite is behaviour-preserving — it matches the old
`/<[^>]*>/g` span for span, byte for byte, including on ragged input:

| input | old | new |
|---|---|---|
| `<scr<script>ipt>` | `ipt>` | `ipt>` |
| `<<script>script>` | `script>` | `script>` |
| `<<a>script>` | `script>` | `script>` |

So the weakness the alert describes is exactly as present as it was before. What
changed is only the *detection*: CodeQL pattern-matches the regex shape, and an
index scan is not a shape it recognises as a sanitizer, so it stopped reporting.
The new alert this PR briefly introduced on its own test file was the proof —
the same rule fired on the test's oracle, the last place the old construct
survived.

A tag stripper cannot be made safe by rewriting the scan. The honest fix for
this alert class is to say so, which is what this PR now does:
`stripHtmlTags`'s JSDoc carries a "not a sanitizer" caveat matching the one
already on `security.sanitizeString`, and `containsHtml` gets a matching note
that it is a heuristic, not a security gate.

The test that pins the old/new equivalence now hardcodes its expected outputs
instead of computing them with `/<[^>]*>/g` — the equivalence claim is worth
testing, but not with the construct being retired.

## Deliberately left (4 alerts, all on `security.sanitizeString`)

#1 `js/bad-tag-filter`, #4 and #5 `js/incomplete-multi-character-sanitization`,
and #11 `js/polynomial-redos`, all on `src/security/index.ts:19`. The issue is
explicit that keep-and-rename vs drop-for-DOMPurify vs dismiss-with-rationale is
the owner's call, not an agent's, so `sanitizeString` is untouched.

**Merger, note:** `Closes #190` will close that issue on merge, so those four
alerts need a follow-up issue or they become invisible.

(The issue's count of 15 included three alerts in
`scripts/generate-module-docs.mjs` that are already closed; 12 were open when
this PR opened.)

## Verification

`pnpm typecheck`, `pnpm check`, `pnpm test` (368 passed, 3 skipped) all clean.
`pnpm docs:generate` produces no diff.

### DIFF
--- a/src/currency/index.test.ts
+++ b/src/currency/index.test.ts
@@ -84,6 +84,7 @@ describe('currency module', () => {
 
 		it('parses European format price strings', () => {
 			expect(parsePrice('€1.234,56')).toBeCloseTo(1234.56)
+			expect(parsePrice('€1.234.567,89')).toBeCloseTo(1234567.89)
 		})
 
 		it('returns null for invalid strings', () => {
--- a/src/currency/index.ts
+++ b/src/currency/index.ts
@@ -182,7 +182,9 @@ export function formatPriceCompact(
  */
 export function parsePrice(value: string): number | null {
 	// Handle European format (1.234,56) by detecting comma as decimal separator
-	const hasEuropeanFormat = /\d+\.\d{3},\d{2}$/.test(value)
+	// One leading \d, not \d+ — the test is the same (both need a digit before the dot), but
+	// with no unbounded quantifier it stays linear on '999…9' (CodeQL js/polynomial-redos).
+	const hasEuropeanFormat = /\d\.\d{3},\d{2}$/.test(value)
 	let normalized = value
 
 	if (hasEuropeanFormat) {
--- a/src/emails/index.test.ts
+++ b/src/emails/index.test.ts
@@ -16,6 +16,12 @@ describe('emails module', () => {
 		expect(isValidEmail('foo@.com')).toBe(false)
 	})
 
+	it('isValidEmail rejects empty and trailing domain labels', () => {
+		expect(isValidEmail('foo@bar..com')).toBe(false)
+		expect(isValidEmail('foo@bar.com.')).toBe(false)
+		expect(isValidEmail('foo.bar@baz.example.com')).toBe(true)
+	})
+
 	it('normalizeEmail trims and lowercases', () => {
 		expect(normalizeEmail('  TEST@Example.COM  ')).toBe('test@example.com')
 	})
--- a/src/emails/index.ts
+++ b/src/emails/index.ts
@@ -12,7 +12,9 @@
  * @returns True if valid, false otherwise.
  */
 export function isValidEmail(email: string): boolean {
-	return /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(email)
+	// The domain parts exclude '.' so each one has a single possible split, which keeps the
+	// match linear on inputs like 'a@!.!.!.!.' (CodeQL js/polynomial-redos).
+	return /^[^\s@]+@[^\s@.]+(?:\.[^\s@.]+)+$/.test(email)
 }
 
 /**
--- a/src/html/index.test.ts
+++ b/src/html/index.test.ts
@@ -21,6 +21,20 @@ describe('html module', () => {
 		expect(stripHtmlTags('no tags')).toBe('no tags')
 	})
 
+	it('stripHtmlTags matches what /<[^>]*>/g did on ragged input', () => {
+		const cases = ['<<a>script>', '<scr<x>ipt>', 'a<b', '<<<', '<>', 'a>b<c>d', '']
+		for (const input of cases) {
+			expect(stripHtmlTags(input)).toBe(input.replace(/<[^>]*>/g, ''))
+		}
+	})
+
+	it('stripHtmlTags leaves nothing a second pass would strip', () => {
+		for (const input of ['<<a>script>', '<scr<x>ipt>', '<a<b>c>']) {
+			const once = stripHtmlTags(input)
+			expect(stripHtmlTags(once)).toBe(once)
+		}
+	})
+
 	it('textToHtml converts newlines to <br>', () => {
 		expect(textToHtml('a\nb\nc')).toBe('a<br>b<br>c')
 		expect(textToHtml('no newlines')).toBe('no newlines')
@@ -30,5 +44,12 @@ describe('html module', () => {
 		expect(containsHtml('<div>')).toBe(true)
 		expect(containsHtml('plain')).toBe(false)
 		expect(containsHtml('a <b>bold</b> word')).toBe(true)
+	})
+
+	it('containsHtml matches what /<[^>]+>/ did on ragged input', () => {
+		const cases = ['<>', '<><a>', '<<>', '<a<b>', 'a<b', '<<<', '']
+		for (const input of cases) {
+			expect(containsHtml(input)).toBe(/<[^>]+>/.test(input))
+		}
 	})
 })
--- a/src/html/index.test.ts
+++ b/src/html/index.test.ts
@@ -21,10 +21,25 @@ describe('html module', () => {
 		expect(stripHtmlTags('no tags')).toBe('no tags')
 	})
 
-	it('stripHtmlTags matches what /<[^>]*>/g did on ragged input', () => {
-		const cases = ['<<a>script>', '<scr<x>ipt>', 'a<b', '<<<', '<>', 'a>b<c>d', '']
-		for (const input of cases) {
-			expect(stripHtmlTags(input)).toBe(input.replace(/<[^>]*>/g, ''))
+	// Expectations are hardcoded, not computed from the retired regex: the old
+	// construct is exactly what this rewrite is removing from the codebase.
+	// Each pair below was checked against the old behaviour by hand.
+	it('stripHtmlTags matches the old regex output on ragged input', () => {
+		const cases: [input: string, expected: string][] = [
+			['<<a>script>', 'script>'],
+			['<scr<x>ipt>', 'ipt>'],
+			['<scr<script>ipt>', 'ipt>'],
+			['<<script>script>', 'script>'],
+			['<scr<x>ipt>alert(1)', 'ipt>alert(1)'],
+			['<img/onerror=x>', ''],
+			['a<b', 'a<b'],
+			['<<<', '<<<'],
+			['<>', ''],
+			['a>b<c>d', 'a>bd'],
+			['', ''],
+		]
+		for (const [input, expected] of cases) {
+			expect(stripHtmlTags(input)).toBe(expected)
 		}
 	})
 
--- a/src/html/index.ts
+++ b/src/html/index.ts
@@ -73,7 +73,17 @@ function findTag(str: string, from: number): { open: number; close: number } | n
  * ```typescript
  * stripHtmlTags('<p>Hello <b>world</b></p>') // 'Hello world'
  * stripHtmlTags('plain text') // 'plain text'
+ *
+ * // Strips `<…>` spans, so ragged input leaves residue — not a sanitizer.
+ * stripHtmlTags('<<a>script>') // 'script>'
+ * stripHtmlTags('<scr<x>ipt>') // 'ipt>'
  * ```
+ *
+ * Removes `<…>` spans and nothing else: it does not decode entities and does
+ * not understand attributes or quoting, so ragged or nested markup leaves
+ * attacker-controlled residue behind rather than a safe string. Not a
+ * substitute for a real sanitizer such as DOMPurify when rendering untrusted
+ * HTML — this produces display text, it does not make untrusted markup safe.
  *
  * @param str The string to strip tags from.
  * @returns The plain text string.
@@ -99,6 +109,11 @@ export function textToHtml(str: string): string {
 
 /**
  * Checks if a string contains any HTML tags.
+ *
+ * A heuristic for "does this look like markup", not a security gate: it only
+ * looks for a non-empty `<…>` span, so `'a<b'` and `'<>'` are both false. Do
+ * not use it to decide whether untrusted input is safe to render.
+ *
  * @param str The string to check.
  * @returns True if the string contains HTML tags, false otherwise.
  */
--- a/src/html/index.ts
+++ b/src/html/index.ts
@@ -51,6 +51,22 @@ export function unescapeHtml(str: string): string {
 }
 
 /**
+ * Finds the next `<…>` span at or after `from`, matching what `/<[^>]*>/` would match.
+ *
+ * ponytail: an index scan rather than the regex — same spans, but linear in the length of
+ * the input instead of quadratic on inputs like `'<<<<<<<'` (CodeQL `js/polynomial-redos`).
+ *
+ * A `<` with no `>` after it is not a span, and neither is anything later in the string,
+ * so `null` ends the search.
+ */
+function findTag(str: string, from: number): { open: number; close: number } | null {
+	const open = str.indexOf('<', from)
+	if (open === -1) return null
+	const close = str.indexOf('>', open + 1)
+	return close === -1 ? null : { open, close }
+}
+
+/**
  * Strips all HTML tags from a string.
  *
  * @example
@@ -63,7 +79,13 @@ export function unescapeHtml(str: string): string {
  * @returns The plain text string.
  */
 export function stripHtmlTags(str: string): string {
-	return str.replace(/<[^>]*>/g, '')
+	let out = ''
+	let i = 0
+	for (let tag = findTag(str, i); tag; tag = findTag(str, i)) {
+		out += str.slice(i, tag.open)
+		i = tag.close + 1
+	}
+	return out + str.slice(i)
 }
 
 /**
@@ -81,5 +103,9 @@ export function textToHtml(str: string): string {
  * @returns True if the string contains HTML tags, false otherwise.
  */
 export function containsHtml(str: string): boolean {
-	return /<[^>]+>/.test(str)
+	// Same matches as /<[^>]+>/ — the `close > open + 1` check is the `+` (an empty `<>` is not a tag).
+	for (let tag = findTag(str, 0); tag; tag = findTag(str, tag.close + 1)) {
+		if (tag.close > tag.open + 1) return true
+	}
+	return false
 }
--- a/src/url/index.test.ts
+++ b/src/url/index.test.ts
@@ -54,6 +54,12 @@ describe('joinUrl', () => {
 		expect(joinUrl('/a/', '/b/', '/c/')).toBe('/a/b/c')
 		expect(joinUrl('a', 'b', 'c')).toBe('a/b/c')
 	})
+
+	it('collapses runs of slashes and drops all-slash segments', () => {
+		expect(joinUrl('https://a.com///', '///foo///', '//bar//')).toBe('https://a.com/foo/bar')
+		expect(joinUrl('a', '///', 'b')).toBe('a/b')
+		expect(joinUrl('///')).toBe('')
+	})
 })
 
 describe('getHostname', () => {
--- a/src/url/index.ts
+++ b/src/url/index.ts
@@ -87,16 +87,27 @@ export function removeQueryParam(url: string, key: string): string {
 }
 
 /**
+ * Trims slashes off a URL segment — always trailing, and leading too when `leading` is set.
+ *
+ * ponytail: an index scan rather than `/^\/+|\/+$/g` — same result, but linear on inputs
+ * like `'//////'` instead of quadratic (CodeQL `js/polynomial-redos`).
+ */
+function trimSlashes(part: string, leading: boolean): string {
+	let start = 0
+	let end = part.length
+	if (leading) while (start < end && part[start] === '/') start++
+	while (end > start && part[end - 1] === '/') end--
+	return part.slice(start, end)
+}
+
+/**
  * Joins multiple URL segments into a single URL, ensuring proper slashes.
  * @param parts The URL segments.
  * @returns The joined URL string.
  */
 export function joinUrl(...parts: string[]): string {
 	return parts
-		.map((part, i) => {
-			if (i === 0) return part.replace(/\/+$/, '')
-			return part.replace(/^\/+|\/+$/g, '')
-		})
+		.map((part, i) => trimSlashes(part, i > 0))
 		.filter(Boolean)
 		.join('/')
 }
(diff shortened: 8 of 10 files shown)


rtorcato (commented):

**Verdict: clean. No defects. `ai-ok-sec`.**

## 1. ReDoS fixes — verified genuinely linear, not moved

- **`emails.isValidEmail` / `validation.isEmail`**: `[^\s@]+@[^\s@]+\.[^\s@]+$` → `[^\s@]+@[^\s@.]+(?:\.[^\s@.]+)+$`. This is the textbook correct rewrite: excluding `.` from the domain-label character class removes the ambiguity between "which literal dot is *the* separator" that made the original polynomial. The delimiter (`.`) in `(?:\.[^\s@.]+)+` is disjoint from the repeated class, so each iteration boundary is forced — no nested-quantifier ambiguity reintroduced. Traced several adversarial inputs (`a@bbbb...` with no dot, `a@` + alternating `b.` runs, multi-`@` strings) — all resolve in O(1) or O(n) with no backtracking blowup.
- **`url.joinUrl`**: regex replaced with a manual two-pointer `trimSlashes` scan. Straightforward, no regex at all, so no ReDoS surface. Verified the leading/trailing trim logic against the original `i===0` vs `i>0` branching — behavior-preserving (checked against the new `///`, `a///b` test cases by hand).
- **`currency.parsePrice`**: `\d+\.` → `\d\.`. Worth noting this one probably wasn't exploitable to begin with (`\d+` immediately before a disjoint literal `.` doesn't have the ambiguity that causes polynomial blowup) — CodeQL likely flagged it on the "any unbounded quantifier reachable from input" heuristic rather than a real quadratic case. The fix is harmless and semantics-preserving either way, not worth blocking on.

## 2. `html.stripHtmlTags` / `containsHtml` — genuinely improved, not differently bypassable

Traced the `findTag` two-pointer scan against the classic reconstruction bypass (`<scr<script>ipt>` style inputs) by hand: because the scan always consumes from an unmatched `<` through to the *first* `>` found anywhere after it (never re-using a `>` already claimed by an earlier span, and never emitting a `<` into `out` unless it's proven to have no later `>` at all), the output can never contain a reconstructible `<...>` pair. That's a real invariant, not just an artifact of the specific test cases — I independently re-derived it rather than trusting the PR's assertion. The equivalence-to-old-regex tests and the second-pass-is-a-no-op test are good evidence for this, not just decoration.

The PR body is appropriately honest that this is "a complexity fix, not a claim that `stripHtmlTags` is a sanitizer" — matches what the code actually does. No overclaiming.

## 3. Scope — residual risk is visible

`security.sanitizeString`'s 4 alerts (`js/bad-tag-filter`, `js/incomplete-multi-character-sanitization` ×2, `js/polynomial-redos`) are correctly left untouched per the issue's explicit "owner's call" framing, and the PR body enumerates them clearly under "Deliberately left" with a note that a follow-up issue is still needed to settle keep/rename/drop. Residual risk is visible, not buried.

No blocking issues. Nit-only: the `currency.parsePrice` alert write-up above, not worth a diff change.

rtorcato (commented):

**Correctness — verified by hand, all five rewrites check out.**

- `currency.parsePrice`: `\d+\.\d{3},\d{2}$` → `\d\.\d{3},\d{2}$`. These two patterns are equivalent as "does a match exist anywhere" predicates against an unanchored `.test()` — if the `\d+` form matches starting at position *p* consuming *k* digits, the single-`\d` form also matches starting at *p+k-1*, and vice versa. No behavior change, ReDoS quantifier genuinely removed.
- `url.joinUrl` / `trimSlashes`: traced both branches (leading=false for the first segment, leading=true otherwise) against the old `/\/+$/` and `/^\/+|\/+$/g` regexes — identical trim boundaries in every case I checked, including the added multi-slash tests.
- `html.stripHtmlTags` / `findTag`: the index-scan greedily finds the same `<…>` span the old `/<[^>]*>/g` would (since `[^>]*` can itself absorb `<`, its greedy extent is bounded only by the next `>`, same as `indexOf`). Walked through `'<<a>script>'` by hand and it matches. The "no reconstruction possible" claim is also provably true: any `<` that survives is either consumed as an `open` by `findTag` (and thus part of a stripped span) or is the terminal `str.slice(i)` after the loop, which by construction has no `>` anywhere after it in the whole string — so a second pass genuinely is a no-op, not just empirically for the tested cases.
- `html.containsHtml`: worth double-checking because the loop advances to `tag.close + 1` on a failed length check rather than `tag.open + 1` — but the only way the check fails is when `close === open + 1` (an empty `<>` body), and in that case there's nothing between `open` and `close` to skip over, so it can't miss a match. Confirmed against the added `'<><a>'` case by hand.
- `emails.isValidEmail` / `validation.isEmail`: the new domain grammar (`[^\s@.]+(?:\.[^\s@.]+)+`) forces every label to be non-empty, correctly rejecting consecutive dots and a trailing dot while still accepting multi-label domains (`baz.example.com`) and everything the existing `foo@bar`, `foo@.com` tests already covered. The rejected set is broader than just the two admitted examples (any consecutive-dot or leading/trailing-dot domain), which the "domain labels now exclude `.`" description in the PR body does accurately convey.

**On the `fix:` vs. breaking-change question (explicitly flagged for a human, not blocking):** the two admitted cases (`foo@bar..com`, `foo@bar.com.`) are arguably genuine bugs — RFC 5321 domain grammar requires non-empty sub-domains, so the old regex was accepting syntactically invalid addresses. That leans toward `fix:` being honest. But this is a published, `exports`-registered validator whose boolean return value now flips for real inputs some downstream consumer may already depend on, and this repo's `semantic-release` treats `fix:` as an unannounced patch bump with no changelog signal beyond "fix". Given the squash-merge title is literally `fix: make the ReDoS-flagged regexes linear-time`, worth the maintainer's explicit call on whether a `BREAKING CHANGE:` footer (or `fix!:`) is warranted so consumers seeing a patch bump aren't surprised by a validator that now rejects input it used to accept. Not something I'd block on — it's a policy call, not a defect, and the PR body already surfaces the tradeoff transparently for review.

**Test coverage — thorough for behavior-equivalence, one real gap.** The html/url/email tests do direct comparisons against the old regex on ragged/adversarial input (`'<<a>script>'`, `'<scr<x>ipt>'`, `'<<<'`, etc.), which is exactly the right way to pin span-for-span equivalence. What's missing across all five locations: no timing/pathological-input regression test (e.g., a long crafted string asserted to complete under some threshold) that would fail if a future edit reintroduced a quadratic pattern. For `html`/`url` this matters less since those are now plain index scans, linear by construction, not regex-based at all. For `currency`/`emails`/`validation`, which are still regex-based rewrites, a correctness-only test wouldn't catch someone later "simplifying" the pattern back into a vulnerable shape. Minor, not blocking.

**Convention check:** module-level helpers (`findTag`, `trimSlashes`) use `function` declarations per `CLAUDE.md`'s stated convention, and the `security.sanitizeString` alerts are correctly left untouched per issue #190's explicit instruction that the sanitizer-owner decision belongs to a human, not an agent.

No defects found. Nice, careful work translating the html regexes into index scans — the containsHtml edge-case reasoning in particular is not obvious and is verified correct.

rtorcato:
🤖 *Automated — `ai-issue-loop` Pass 1 (handoff, retrofitted). Posted under the owner's account by an agent; not a human message. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

This PR passed both reviewers, but the code review left findings that the label
state alone did not carry. The `ai-notes` label and the closing section below are
new — this comment backfills them for a PR reviewed before they existed. Future
reviews will end with this section themselves.

### Before merging

- **`fix:` may understate this.** `isValidEmail` / `isEmail` now reject
  `foo@bar..com` and `foo@bar.com.` — input previously accepted. semantic-release
  publishes off the squash subject, so this decides whether consumers get a patch
  or a major. Worth a deliberate call rather than inheriting the default.
- **No timing test for the pathological input.** The ReDoS fixes have no
  regression test that would fail if the backtracking came back, so nothing stops
  a later refactor from silently reintroducing it.
- **4 alerts stay open by design.** `security.sanitizeString`'s
  `js/incomplete-multi-character-sanitization` ×2, `js/bad-tag-filter`, and
  `js/polynomial-redos` were deliberately left alone because #190 says keep /
  rename / drop is your call. `Closes #190` will close that issue on merge, so
  those four need a follow-up issue or they become invisible.

rtorcato:
🤖 *Automated — `ai-issue-loop` Pass 1. Posted under the owner's account by an agent; not a human message. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

Both reviewers passed this PR, but it **cannot merge** — so it is going back for
one fix round. The reviewers are diff-scoped and do not run CodeQL, which is why
neither caught this.

## The blocker

This PR introduces a new high-severity alert **in a test file it added**:

```
src/html/index.test.ts:27
js/incomplete-multi-character-sanitization  (high)
"This string may still contain <script, which may cause an HTML element
 injection vulnerability."
```

The `code-scanning-main` ruleset is active on `main` with
`security_alerts_threshold: high_or_higher`, so GitHub refuses the merge while
that alert exists. `mergeStateStatus` is `BLOCKED`. All five *required* checks
(`lint`, `typecheck`, `build`, `test (22)`, `test (24)`) pass — this is the
ruleset, not a required check.

Worth being clear that the PR is a net win and is not being rejected: it takes
`main` from **12 open high alerts to 1**, and that 1 is this new one.

## What to fix

Only the test fixture at `src/html/index.test.ts:27`. Rewrite the input string
so it still exercises `stripHtmlTags` meaningfully without CodeQL reading the
test itself as a sanitizer — e.g. build the payload from concatenated fragments,
or assert against a fixture that does not embed a literal partial `<script`.

**Do not weaken the production fix in `src/html/index.ts` to make the alert go
away** — the shipped code is what the reviewers verified by hand.

## Explicitly out of scope

`src/security/index.ts`'s 4 alerts on `sanitizeString` stay untouched. Issue #190
says keep / rename / drop is the owner's call, and that decision has not been
made. Leave them.

Note for whoever merges this later: `Closes #190` will close that issue, so those
4 alerts need a follow-up issue or they become invisible.

rtorcato:
🤖 *Automated — `ai-issue-loop` Pass 1 (correction to my previous comment). Posted under the owner's account by an agent; not a human message. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

**Correcting myself.** My previous comment said to fix only the test fixture at
`src/html/index.test.ts:27`. That would clear the alert and hide a real finding.
Do not just rewrite the test.

## What the alert is actually pointing at

Alert #33 fires on the test's *oracle*, not on a fixture string:

```ts
expect(stripHtmlTags(input)).toBe(input.replace(/<[^>]*>/g, ''))
```

The test asserts the new `stripHtmlTags` is **exactly equivalent** to the old
`/<[^>]*>/g`. It is — verified on the ragged inputs, byte-identical:

| input | old | new | same |
|---|---|---|---|
| `<scr<script>ipt>` | `ipt>` | `ipt>` | ✅ |
| `<<script>script>` | `script>` | `script>` | ✅ |
| `<<a>script>` | `script>` | `script>` | ✅ |

So the index-scan rewrite is a **performance change, not a security change**. The
`js/polynomial-redos` fix is real and good — quadratic to linear. But
`js/incomplete-multi-character-sanitization` on `stripHtmlTags` was not fixed. It
stopped being reported because CodeQL no longer recognises an index scan as a
sanitizer; the regex shape it pattern-matches on is gone. The function still
leaves `script>` and `ipt>` residue exactly as before.

The new alert is the proof: the same rule now fires on the test, because the test
is the only place the old construct survives. CodeQL is right both times — it
just cannot see the production version any more.

## What to do instead

1. **Add the caveat to `stripHtmlTags`'s JSDoc.** `security.sanitizeString`
   already carries one ("Blocklist-based, so it is bypassable — not a substitute
   for a real sanitizer such as DOMPurify"). `stripHtmlTags` carries none, and it
   is bypassable in the same way. That is the honest fix for this alert class —
   a tag stripper cannot be made safe by rewriting the scan; it gets documented.
2. **Then** rewrite the test oracle so it does not embed
   `.replace(/<[^>]*>/g, '')` — hardcode the expected outputs from the table
   above. The equivalence claim is still worth testing; it just should not be
   computed with the construct being retired.
3. **Correct the PR body.** It states this PR fixed the `stripHtmlTags`
   incomplete-sanitization alert. It did not — it silenced it. Say that plainly
   so the next reader is not misled.

Still out of scope, unchanged: `src/security/index.ts`'s 4 alerts on
`sanitizeString`. Issue #190 says keep / rename / drop is the owner's call.

rtorcato (commented):

Re-review after the fix round. Verdict: clean.

**1. `stripHtmlTags` JSDoc caveat — accurate and sufficient.**
The new doc comment states plainly that the function "removes `<…>` spans and nothing else," does not decode entities or understand attributes/quoting, and that "ragged or nested markup leaves attacker-controlled residue behind rather than a safe string." It includes concrete examples in the doc block itself (`stripHtmlTags('<<a>script>') // 'script>'`). This is at least as strong as the existing `security.sanitizeString` caveat ("Blocklist-based, so it is bypassable — not a substitute for a real sanitizer such as DOMPurify") and doesn't understate the risk. `containsHtml` also picked up a correct "heuristic, not a security gate" note. No under-statement found.

**2. PR body honesty — correct.**
The body explicitly retracts the earlier claim ("Correcting an earlier version of this PR description, which claimed this alert was fixed. It was not.") and states the weakness "is exactly as present as it was before" — only CodeQL's detection changed, not the underlying behavior. The old-regex-vs-new-scan equivalence table matches what I'd expect from an `indexOf`-based span scan replicating `/<[^>]*>/g`. This is honest.

**3. ReDoS fixes — still genuinely linear, this round didn't disturb them.**
- `emails.isValidEmail` / `validation.isEmail`: `[^\s@]+@[^\s@]+\.[^\s@]+` → `[^\s@]+@[^\s@.]+(?:\.[^\s@.]+)+`. Excluding `.` from the domain char classes removes the split ambiguity that caused backtracking; each label position is now unique. Semantics check out against the new tests (`bar..com` / `bar.com.` rejected, `baz.example.com` still accepted).
- `currency.parsePrice`: `\d+\.\d{3},\d{2}$` → `\d\.\d{3},\d{2}$`. Since the match is unanchored at the start, requiring exactly one digit before the dot vs. one-or-more is behaviorally equivalent for this boolean test (a digit run before the dot either satisfies both or neither), and removing the `+` eliminates the backtracking source. New test (`€1.234.567,89`) confirms no regression.
- `url.joinUrl` (`trimSlashes`) and `html.stripHtmlTags`/`containsHtml` (`findTag`): both are single-pass index scans with forward-only cursor advance — no backtracking possible, straightforwardly O(n).

Nothing here reintroduces the quadratic shape; the fixes look real.

**4. Residual risk visibility — confirmed.**
The PR body's "Deliberately left" section explicitly flags: "**Merger, note:** `Closes #190` will close that issue on merge, so those four alerts need a follow-up issue or they become invisible." That's exactly the disclosure #190 asked for, and it correctly scopes the sanitizeString decision as the owner's call, not this PR's.

### Before merging
File the follow-up issue for the 4 `sanitizeString` alerts (`js/incomplete-multi-character-sanitization` ×2, `js/bad-tag-filter`, `js/polynomial-redos`) before or immediately after merging — `Closes #190` will otherwise make that residual risk invisible once this closes.

rtorcato (commented):

Reviewed the current diff against the three things this fix round was meant to change.

**1. "Not a sanitizer" caveats (`stripHtmlTags` JSDoc, `containsHtml` note).**
Traced both by hand against the actual `findTag`/`stripHtmlTags`/`containsHtml` implementation rather than taking the prose on faith:

- `stripHtmlTags`'s caveat ("removes `<…>` spans and nothing else... ragged or nested markup leaves attacker-controlled residue... not a substitute for a real sanitizer such as DOMPurify") matches the actual behaviour — verified `<<a>script>` → `'script>'`, `<scr<x>ipt>` → `'ipt>'` by tracing `findTag`'s `indexOf('<')`/`indexOf('>')` scan step by step. Neither over- nor underclaims: it doesn't say "unsafe" categorically, just names the specific failure mode (residue on ragged input), which is accurate and matches the sibling caveat's style on `security.sanitizeString` described in #190.
- `containsHtml`'s note ("only looks for a non-empty `<…>` span, so `'a<b'` and `'<>'` are both false") is also correct: `'a<b'` has no `>` so `findTag` returns `null` immediately; `'<>'` finds `open=0, close=1`, and `close > open + 1` is `1 > 1` = false. Both claims check out.

**2. Hardcoded `[input, expected]` table in `src/html/index.test.ts`.**
This is the part with real risk (a wrong hardcoded value silently matching a buggy implementation), so I traced all 11 cases through `findTag`/`stripHtmlTags` directly rather than trusting the table:

`<<a>script>`→`script>`, `<scr<x>ipt>`→`ipt>`, `<scr<script>ipt>`→`ipt>`, `<<script>script>`→`script>`, `<scr<x>ipt>alert(1)`→`ipt>alert(1)`, ``→`''`, `a<b`→`a<b`, `<<<`→`<<<`, `<>`→`''`, `a>b<c>d`→`a>bd`, `''`→`''`.

All 11 are correct — every one matches what a hand-trace of the `indexOf`-based scan produces, and the three that overlap with the PR description's table (`<<a>script>`, `<scr<script>ipt>`, `<<script>script>`) agree with it. The `containsHtml` equivalence test still compares live against `/<[^>]+>/`, so no hardcoding risk there. The idempotency test (`stripHtmlTags(once) === stripHtmlTags(stripHtmlTags(once))`) is a legitimate regression guard, though for the three sample inputs the second pass is a no-op (no `<` survives the first pass) rather than exercising a case where a second pass would actually strip something — a minor gap, not a defect. This change also plausibly explains the prior CodeQL block: the removed `.replace(/<[^>]*>/g, '')` oracle is gone from the test file entirely, and nothing in the new table reintroduces a regex-shaped construct CodeQL would read as an incomplete sanitizer.

**3. No production behaviour change this round.**
`gh pr diff` only gives the cumulative diff against `main`, not a round-scoped diff, so this can't be fully isolated — but nothing in `src/html/index.ts` beyond JSDoc/comment additions is inconsistent with a docs-only round: `findTag`, `stripHtmlTags`, and `containsHtml`'s logic (indexOf scan, slicing, the `close > open + 1` heuristic check) are exactly what round 1's index-scan rewrite would produce, with no new branches or altered control flow. Combined with the prior reviewer comment's explicit instruction to touch only the test fixture, I have no basis to suspect a behaviour change slipped in here.

**Carried forward, unresolved:** `isValidEmail`/`isEmail` now reject `foo@bar..com` and `foo@bar.com.`, which were previously accepted — that's a behaviour change for existing consumers, and this PR is titled `fix:`. semantic-release publishes off the squash subject, so whether this ships as a patch or is called out as breaking is a real decision, not a formality. Practically, `feat(modules)!` is already unreleased on `main` with its own BREAKING CHANGE footer, so the *next release is 3.0.0 regardless* of what this PR's commit type says — this doesn't gate the version number. It still affects what the changelog tells consumers about this specific behaviour change, so it's worth a deliberate call by whoever merges, not a silent inheritance of `fix:`. This is a judgment call, not something an agent should resolve.

### Before merging
- Decide whether the `isValidEmail`/`isEmail` behaviour change (rejecting `foo@bar..com` and `foo@bar.com.`) should be called out as breaking in the squash commit/changelog, even though the next release is 3.0.0 regardless.

Issue #185, opened 2026-08-12
Docs advertise exports that do not exist


rtorcato:
Found while executing #174. Several user-facing code samples import names the package has never exported. The landing-page quick start is among them, so the *first* thing a new user copy-pastes fails to build.

## Confirmed phantom names

Checked against `src/*/index.ts` on `main`:

| Advertised | Actually exported |
|---|---|
| `uuid.generateUUID` | `getUUID` |
| `uuid.isValidUUID` | `isUUID` |
| `strings.removeEmojis` | — nothing like it exists |
| `emails.validateEmail` | `isValidEmail` |
| `promises.timeout` | `withTimeout` |
| `promises.retry` | — nothing like it exists |

## Where they appear

- `README.md:117,125` — "Library Usage" block
- `apps/docs/docs/index.mdx:48,55` — **quick start**, the highest-traffic sample
- `apps/docs/docs/index.mdx:84,86,87` — category table
- `apps/docs/docs/modules/overview.md:16,20,22,25`
- `apps/docs/docs/modules/uuid.md:10,12,15,16`
- `apps/docs/docs/modules/strings.md:10,15`

## Why it slipped through

`scripts/check-readme-exports.mjs` only validates import samples inside the README's `## Available Modules` section. `README.md:117` sits in `## Library Usage`, above it, and the docs site is not checked at all.

## Scope

Mechanical: replace each phantom name with the real export and fix the sample's output comment to match actual behaviour (e.g. `getUUID()` vs `generateUUID()`). Verify by grepping the replacement against the module's `index.ts` — do not guess a name that merely looks plausible.

Do **not** add the missing helpers (`removeEmojis`, `retry`) as part of this. Adding API is a separate call and the module paths are frozen as of #174.

Closing the enforcement gap is tracked separately — see the follow-up issue on the export checker, which should land after this one.

PR #198, opened 2026-08-12
docs: replace phantom export names with the real ones

rtorcato:
The quick start, README usage block and several module pages imported
names the package has never exported, so the first sample a new user
copy-pastes fails to build.

Verified every replacement against src/<module>/index.ts:

- uuid.generateUUID -> getUUID, uuid.isValidUUID -> isUUID
- emails.validateEmail -> isValidEmail (plus normalizeEmail, maskEmail)
- promises.timeout -> withTimeout; promises.retry does not exist -> to
- strings.removeEmojis does not exist -> pluralize / titleCase / escapeHtml
- numbers.formatCurrency -> currency.formatPrice
- the `math` subpath does not exist -> geometry

No helpers were added: the missing removeEmojis and retry are a separate
API call, and the module paths are frozen as of #174.

Closes #185

PR #199, opened 2026-08-12
ci(boundaries): enforce the module-boundary freeze in CI

### LINKED ISSUE #187: Nothing enforces the module-boundary freeze in CI
Follow-up to #174. `MODULE-BOUNDARIES.md` freezes module paths and says every export name has exactly one home — but nothing checks it. The doc admits this itself:

> `scripts/check-readme-exports.mjs` catches drift between the README and `package.json`; it does not catch a name exported from two modules, so that check stays a review responsibility.

A freeze defended only by reviewer vigilance is the kind that quietly stops holding. The whole point of #174 was that re-introducing a duplicate is free to fix during beta and expensive after 1.0, so the guard is worth more now than it will be later.

## Two gaps

**1. No duplicate-name check.** Nothing fails when a helper is exported from two subpath modules. This is the exact condition #174 spent an audit removing — it should be one CI check, not a thing a reviewer has to notice.

**2. The existing checker sees too little.** `scripts/check-readme-exports.mjs` validates import samples only inside the README's `## Available Modules` section. It does not look at the rest of the README, and it does not look at `apps/docs/**` at all — which is how the phantom exports in the sibling docs issue survived, including the one in the docs-site quick start.

## Scope

Extend `scripts/check-readme-exports.mjs` (or add a sibling script wired into the same CI step) to fail on:

- any export name reachable from two different `package.json#exports` subpaths
- any `import { … } from '@rtorcato/js-common/<module>'` sample, anywhere in `README.md` or `apps/docs/docs/**`, naming something that module does not export

Reuse the existing `collectExportNames()` — it already resolves local re-exports.

The `@rtorcato/js-common/types` subpath is types-only and has no `import` field; the walker must not treat that as an error.

## Ordering

**Do this after the phantom-exports issue lands.** Gap 2's check fails against `main` today by design — that is what it is meant to catch — so landing the checker first would red the branch for reasons the checker did not cause. Branch this off that issue's branch if they are in flight together, and rebase onto `main` after it merges.

rtorcato:
MODULE-BOUNDARIES.md froze module paths and the "every export name has
exactly one home" rule, then admitted nothing checked it — the duplicate
check was left to reviewer vigilance, which is the kind of guard that
quietly stops holding.

Extend scripts/check-readme-exports.mjs (already wired into CI) with the
two missing checks:

- fail when an export name is reachable from two different
  package.json#exports subpaths
- validate every `import { … } from '@rtorcato/js-common/<module>'` sample
  anywhere in README.md or apps/docs/docs/**, not just inside the README's
  `## Available Modules` section

Both reuse the existing collectExportNames(), so local re-exports resolve.
`./types` is declaration-only and has no src/types/index.ts — it is treated
as absent rather than broken.

A migration guide has to quote the API it is migrating away from, so a
fenced block containing `boundary-check: ignore` is skipped. The two
`errors.tryCatch` "before" snippets are the only users.

The success line now prints module, name and file counts so a vacuous pass
(regex matching nothing) is visible rather than silently green.

Gap 2 found four real phantom exports, fixed here so the guard lands green:
generateUUID → getUUID, isValidUUID → isUUID, and removeEmojis → titleCase
(it never existed). #185 remains open for the wider docs audit.

Closes #187

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,10 +104,10 @@ jobs:
             - name: 🤖 Check AGENTS.md is in sync with SKILL.md
               run: node scripts/sync-agents.mjs --check
 
-            # The README API reference is hand-written — fail the PR when a new
-            # module ships without a README entry, or an example imports a name
-            # the module no longer exports.
-            - name: 📚 Check README matches package.json exports
+            # Guards the MODULE-BOUNDARIES.md freeze: a module shipping without a
+            # README entry, a name exported from two subpaths, or a doc example
+            # importing something the module does not export.
+            - name: 📚 Check module boundaries and doc examples
               run: node scripts/check-readme-exports.mjs --check
 
     commitlint:
--- a/MODULE-BOUNDARIES.md
+++ b/MODULE-BOUNDARIES.md
@@ -139,6 +139,20 @@ imported it, so from here it needs a major version and an entry in
 
 Adding a *new* helper to an existing module is not covered by the freeze and
 stays ordinary work — but it inherits the rule at the top: check that the name is
-not already exported somewhere else first. `scripts/check-readme-exports.mjs`
-catches drift between the README and `package.json`; it does not catch a name
-exported from two modules, so that check stays a review responsibility.
+not already exported somewhere else first.
+
+`scripts/check-readme-exports.mjs` enforces that in CI so it does not rest on
+reviewer vigilance. It fails the build on three things:
+
+1. A `package.json#exports` subpath missing from the README's
+   `## Available Modules` section, or a module listed there that is not exported.
+2. **An export name reachable from two different subpaths** — the rule at the top
+   of this record.
+3. An `import { … } from '@rtorcato/js-common/<module>'` sample anywhere in
+   `README.md` or `apps/docs/docs/**` naming something that module does not
+   export.
+
+Docs that quote a removed API on purpose — the "before" half of a migration
+snippet — opt out by putting `boundary-check: ignore` inside the fenced block.
+Use it only for code that is *meant* to be historical; it is not a way to park a
+broken example.
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ npx skills add https://github.com/rtorcato/js-common --skill js-common
 **1.x → 2.x** — the only breaking change in 2.0 is a rename in the `errors` module:
 
 ```ts
+// boundary-check: ignore — quotes the pre-2.0 API on purpose
 // 1.x — swallow errors and fall back to a default value
 import { tryCatch } from '@rtorcato/js-common/errors'
 const data = await tryCatch(() => fetchData(), [])
@@ -114,7 +115,7 @@ See [CLI.md](./CLI.md) for complete CLI documentation.
 // Import specific modules (recommended for tree-shaking)
 import { formatDate, today, daysBetween } from '@rtorcato/js-common/date'
 import { isValidEmail, normalizeEmail } from '@rtorcato/js-common/emails'
-import { generateUUID, isValidUUID } from '@rtorcato/js-common/uuid'
+import { getUUID, isUUID } from '@rtorcato/js-common/uuid'
 import { sum, average, roundTo } from '@rtorcato/js-common/numbers'
 import { capitalize, titleCase } from '@rtorcato/js-common/strings'
 
@@ -122,7 +123,7 @@ import { capitalize, titleCase } from '@rtorcato/js-common/strings'
 console.log(today()) // "2026-05-29"
 console.log(sum([1, 2, 3, 4, 5])) // 15
 console.log(capitalize('hello world')) // "Hello world"
-console.log(generateUUID()) // "f47ac10b-58cc-4372-a567-0e02b2c3d479"
+console.log(getUUID()) // "f47ac10b-58cc-4372-a567-0e02b2c3d479"
 ```
 
 ### Bundle Size Impact
--- a/apps/docs/docs/guides/migration.md
+++ b/apps/docs/docs/guides/migration.md
@@ -74,6 +74,7 @@ different function from awaiting a DOM event, so the event one moved aside.
 ## 1.x → 2.x — `errors.tryCatch` → `errors.tryWithFallback`
 
 ```ts
+// boundary-check: ignore — quotes the pre-2.0 API on purpose
 // 1.x — swallow errors and fall back to a default value
 import { tryCatch } from '@rtorcato/js-common/errors'
 const data = await tryCatch(() => fetchData(), [])
--- a/apps/docs/docs/index.mdx
+++ b/apps/docs/docs/index.mdx
@@ -45,14 +45,14 @@ Import named functions from a subpath export to keep your bundle tiny:
 import { today, daysBetween } from '@rtorcato/js-common/date'
 import { sum, average } from '@rtorcato/js-common/numbers'
 import { capitalize } from '@rtorcato/js-common/strings'
-import { generateUUID } from '@rtorcato/js-common/uuid'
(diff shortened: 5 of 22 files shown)


rtorcato (commented):

**Verdict: no blocking findings.**

**1. File-walking / regex safety (in scope given #190's `js/polynomial-redos` history)**
- `docFiles` only ever globs a fixed, repo-relative path (`apps/docs/docs/**` via `readdirSync(..., { recursive: true })`) filtered to `.md`/`.mdx`; no user-controlled path segments reach `join()`, so there's no traversal risk.
- `IMPORT_EXAMPLE` (`import\s+(?:type\s+)?\{([^}]+)\}\s*from\s*['"]@rtorcato\/js-common\/([\w-]+)['"]`) has no nested/overlapping quantifiers — linear, not a ReDoS candidate.
- `FENCE = /^```[\s\S]*?^```/gm` is the one worth watching: a lazy `[\s\S]*?` under `/m` is not exponential, but it is quadratic in the pathological case (many unmatched/opening ` ``` ` lines in one file force repeated forward scans). Since `apps/docs/docs/**` is PR-editable content, a contributor could in principle land a large doc file that makes this CI step slow. Impact is limited to wasting the job's own CI time (no RCE, no data exposure) and GitHub Actions job timeouts bound the damage — **low severity, not blocking**, but worth a heads-up since this repo actively tracks CodeQL `js/polynomial-redos` findings and this pattern is a plausible future hit.
- No unbounded reads: `readFileSync` is called only on files already filtered by extension, not sized externally.

**2. `boundary-check: ignore` escape hatch**
- Correctly scoped: the fence-strip only feeds check #3 (import-sample validation against docs). It has zero effect on check #1 (README subpath completeness, computed from `package.json#exports`) and check #2 (the cross-module duplicate-name freeze, computed purely from `src/`). So the escape hatch cannot be used to mask a real "two homes for one export" violation — the core guard this PR/issue #187 exists to add stays unconditional.
- Blast radius is genuinely doc-only, and any use of the tag is visible in the PR diff for human review — this is a documented convention (`MODULE-BOUNDARIES.md`: "not a way to park a broken example"), not a technical guarantee. A careless contributor could tag a fenced block containing an actually-broken (non-migration) import to silence it, since the match is "string present anywhere in the block," not "on the first line only." Low risk given it still requires a reviewed PR to land, but worth naming as the tradeoff it is.

**3. Dependencies / secrets / CI privileges**
- No new/changed dependencies, no secrets or tokens touched.
- The renamed workflow step (`📚 Check README matches package.json exports` → `📚 Check module boundaries and doc examples`) is cosmetic only — same `run: node scripts/check-readme-exports.mjs --check`, same job, no permissions/trigger changes in the diff.
- The four export-name fixes in README/docs (`generateUUID`→`getUUID`, `isValidUUID`→`isUUID`, `removeEmojis`→`titleCase`) are consistent with the PR description (phantom exports) and don't touch runtime code.

### Before merging
Nothing blocking. Optional: consider capping/streaming the `FENCE` regex pass (e.g., process line-by-line instead of `[\s\S]*?`) if CodeQL flags it as `js/polynomial-redos` later — not required for this PR.

rtorcato (commented):

**Correctness / does the check actually fail when it should?**

Check 2 (duplicate-name / one-home-per-export) is sound: builds a `name → [subpaths]` map over `exportsBySubpath` and flags any name with more than one entry. Straightforward, matches the described "inject a duplicate `slugify` export, watch it fail, revert" verification.

Check 3 (import-sample validation) is also logically correct — regex extracts `{ names }` from `@rtorcato/js-common/<module>` imports (including `import type { … }`), compares against `collectExportNames()` for that subpath, and reports which file/name/module. Verified by the phantom-import test described in the PR body.

One gap between the stated goal and the result: the PR body says the success line now prints counts "so a vacuous pass … is visible," but for check 3 the message only reports `docFiles.length` (files scanned), not the number of import samples actually matched. If the `IMPORT_EXAMPLE` regex ever stops matching (quote-style change, CRLF, a new fence style, etc.), the file count stays identical and the check still prints success — exactly the vacuous-pass failure mode the PR set out to make visible. Check 2's `homes.size` count doesn't have this problem since it's derived from real export data, not from regex matches on prose. Worth a follow-up: print the number of import statements actually checked, not just files touched. Not blocking — the matching logic itself is correct and was tested — but the "counts prove non-vacuousness" claim is only half true.

**Escape hatch (`boundary-check: ignore`)**

Scoped correctly: `FENCE = /^```[\s\S]*?^```/gm` is non-greedy, so it only strips the individual fenced block containing the marker, not everything after it — confirmed against the diff, where only the "before" `errors.tryCatch` block is marked and the adjacent "after" block is left checked. Documented in `MODULE-BOUNDARIES.md` with a "not a way to park a broken example" caveat, and currently has exactly two call sites, both genuine historical snippets.

That said, this is a text-string escape hatch with zero enforcement behind the "use it only for historical code" instruction — nothing stops a future PR from sprinkling `boundary-check: ignore` onto a fenced block to silence a real phantom import, and the success message gives no visibility into how many blocks were exempted (unlike the export/name counts, there's no "N blocks skipped" line). Given there are only two uses today this isn't a blocker, but if this hatch starts spreading, the checker should probably start reporting an ignored-block count so a reviewer can eyeball whether skips are growing faster than migration guides are.

**Scope creep — phantom export fixes**

Fixing `generateUUID`→`getUUID`, `isValidUUID`→`isUUID`, `removeEmojis`→`titleCase` in this PR is functionally justified — check 3 as written would otherwise land red against `main`, and issue #187 itself anticipated this dependency. But #187's own "Ordering" section says to do this *after* the phantom-exports issue lands, or branch off that issue's branch if the two are in flight together — not to re-fix the same lines independently. This PR did the latter.

**Before merging**

PR #198 (issue #185) is currently in review and fixes the exact same phantom exports in the exact same files: `README.md`, `apps/docs/docs/index.mdx`, `apps/docs/docs/modules/strings.md`, and `apps/docs/docs/modules/uuid.md`. Whichever of #198 / #199 merges second will go `DIRTY` on the overlapping hunks and need a rebase (or the duplicate doc fix dropped in favor of the other PR's version). Recommend merging one first, then rebasing the other and dropping its now-redundant doc changes before merge — otherwise there's a real risk of re-adding a phantom name or reverting the other PR's fix during conflict resolution.

rtorcato:
🤖 *Automated — `ai-issue-loop` Pass 1. Posted under the owner's account by an agent; not a human message. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

Pinning down the merge order the review left open. It recommended "merging one
first, then rebasing the other" without saying which — read on its own, that is
not actionable.

### Before merging

**Merge order: 198 first, then this PR (199).**

Rationale: 199 only carries the overlapping doc fixes so its own new CI check
could land green. Once 198 is on `main` those hunks are already there, so the
rebase mostly evaporates. The reverse order works but is messier — 198 would be
rebasing onto a `main` that already contains most of its own changes.

The same order is now recorded on 198, so the two PRs agree.

**The conflict-resolution hazard the review raised is the real point**, and it
survives whichever order you pick: both PRs edit the same hunks in `README.md`,
`apps/docs/docs/index.mdx`, `apps/docs/docs/modules/strings.md` and
`apps/docs/docs/modules/uuid.md`, so a careless resolution can re-add a phantom
name or revert the other PR's fix. When rebasing the second one, prefer
`main`'s version of every overlapping doc hunk and keep only what is genuinely
unique to that PR.

Worth noting: **once this PR is merged, that hazard stops being silent.** The
check it adds fails CI on any phantom import reintroduced by a bad conflict
resolution — including one introduced while rebasing this very PR. So if the
rebase goes wrong, CI catches it rather than a reader noticing months later.
That is the argument for landing 199 promptly after 198 rather than leaving it
open.

PR #200, opened 2026-08-12
fix(cli): print importable export names from `add`

### LINKED ISSUE #186: CLI `add` prints an import statement nobody can use
Found while executing #174. The `js-common add` command exists to hand a developer a ready-to-paste import. Every import it emits is currently fiction.

## The bug

`src/cli/cli.ts` builds the printed import from `functionCategories[...].functions[].name`. Those are **CLI verb names**, not library export names:

```
$ js-common add   → pick "🔢 Mathematical" → pick sum, avg, random

import { sum, avg, random } from '@rtorcato/js-common/numbers'
```

`numbers` exports `sum`, but there is no `avg` and no `random` (they are `average`, and randomness lives in `./random`). The same applies across the catalog:

| Printed | Real export | Module |
|---|---|---|
| `avg` | `average` | `numbers` |
| `random` / `randomBetween` | `randomInt` | `random` |
| `round` | `roundTo` | `numbers` |
| `title` / `toTitleCase` | `titleCase` | `strings` |
| `pad` / `padWithZeros` | `padStart` | `strings` |
| `now` | `nowIso` | `datetime` |
| `between` | `daysBetween` | `date` |
| `exists` | `fileExists` | `file` |
| `ext` | `getFileExtension` | `file` |
| `password` / `checkPasswordStrength` | `isStrongPassword` | `security` |
| `token` | `generateSecureToken` | `security` |
| `url` | `isUrl` | `validation` |
| `pid` | `getProcessId` | `process` |
| `uptime` | `getProcessUptime` | `process` |
| `ci` / `isCiEnvironment` | `isCI` | `process` |
| `node-version` / `getNodeVersion` | `getNodeMajorVersion` | `node` |

The wrong names also reach the usage snippets in `generateUsageExample()` (`src/cli/cli.ts` ~line 287).

Note a category's verbs do not all live in one module — `math` spans `numbers` and `random`, `system` spans `process` and `node`. So the fix is per-entry, not per-category.

## Already partly addressed

PR #184 added a `subpath` field per category so the *module* in the printed import is real — it was emitting `@rtorcato/js-common/math` after that module was deleted. The **function names** were left alone deliberately, as a pre-existing defect out of that PR's scope. This issue is that remainder.

## Scope

Give each entry in `functionCategories` an explicit `exportName` and `subpath`, then build both the import statement and the usage example from those. Keep the verb `name` as-is — it is the CLI command (`js-common math avg`) and renaming it would break the CLI surface.

Verify every `exportName` against the module's `src/<module>/index.ts`. A test asserting each catalog entry resolves to a real export would stop this drifting again.

rtorcato:
`js-common add` built its printed import and usage examples from the CLI **verb** names, not the library export names — so every import it emitted was unusable (`avg`, `random`, `title`, `pad`, `now`, `exists`, `ci`, …).

## What changed

- Each catalog entry now carries an explicit `exportName` and `subpath`; the import statement and the usage example are both built from those.
- A category can span modules, so `add` groups the selection by subpath and prints one import line per module — e.g. picking `today`, `now`, `between` now prints:

  ```
  import { today, daysBetween } from '@rtorcato/js-common/date'
  import { nowIso } from '@rtorcato/js-common/datetime'
  ```

- The catalog moved to `src/cli/catalog.ts` so `src/cli/catalog.test.ts` can assert every `exportName` really is exported by `src/<subpath>/index.ts` (20 entries, all green). That is the guard against future drift.
- `generateUsageExample()` deleted — each example now lives on the entry next to the name it uses, so the two can't disagree.

## Verified

- `tsc --noEmit`, `biome check .`, `vitest run` (402 passed) all clean.
- `pnpm run build-cli` emits `dist/cli/catalog.js` alongside `cli.mjs` and the binary runs (`list`, `date today`, and `add` driven through a pty to confirm the grouped import output above). Package is `type: module`, so the extra emitted `.js` resolves as ESM without touching the build script.

Closes #186

### DIFF
--- a/src/cli/catalog.test.ts
+++ b/src/cli/catalog.test.ts
@@ -1,0 +1,17 @@
+import { describe, expect, it } from 'vitest'
+import { functionCategories } from './catalog.js'
+
+const entries = Object.entries(functionCategories).flatMap(([category, { functions }]) =>
+	functions.map((func) => ({ category, ...func }))
+)
+
+describe('cli function catalog', () => {
+	it.each(entries)(
+		'js-common $category $name → $exportName from ./$subpath',
+		async ({ exportName, subpath }) => {
+			const mod = await import(`../${subpath}/index.ts`)
+
+			expect(Object.keys(mod)).toContain(exportName)
+		}
+	)
+})
--- a/src/cli/catalog.ts
+++ b/src/cli/catalog.ts
@@ -1,0 +1,194 @@
+// Catalog of the functions the CLI advertises, used by `list`, `interactive`
+// and `add`.
+//
+// Three names per entry, and they are deliberately different things:
+//   - `name`       the CLI verb (`js-common math avg`) — part of the CLI surface
+//   - `exportName` the real library export (`average`) — what `add` must print
+//   - `subpath`    the published module the export lives in (`numbers`)
+//
+// `subpath` is per entry, not per category: the CLI groups by task ('math',
+// 'system') while the library groups by module — 'math' spans `numbers` and
+// `random`, 'system' spans `process` and `node`.
+//
+// `src/cli/catalog.test.ts` asserts every `exportName` really is exported by
+// `src/<subpath>/index.ts`, so this cannot drift back into fiction.
+export const functionCategories = {
+	date: {
+		name: '📅 Date & Time',
+		functions: [
+			{
+				name: 'today',
+				description: "Get today's date (YYYY-MM-DD)",
+				exportName: 'today',
+				subpath: 'date',
+				example: `const date = today() // '${new Date().toISOString().slice(0, 10)}'`,
+			},
+			{
+				name: 'now',
+				description: 'Get current timestamp',
+				exportName: 'nowIso',
+				subpath: 'datetime',
+				example: `const timestamp = nowIso() // '${new Date().toISOString()}'`,
+			},
+			{
+				name: 'between',
+				description: 'Calculate days between dates',
+				exportName: 'daysBetween',
+				subpath: 'date',
+				example: `const days = daysBetween('2023-01-01', '2023-12-31') // 364`,
+			},
+		],
+	},
+	math: {
+		name: '🔢 Mathematical',
+		functions: [
+			{
+				name: 'sum',
+				description: 'Calculate sum of numbers',
+				exportName: 'sum',
+				subpath: 'numbers',
+				example: `const total = sum([1, 2, 3, 4, 5]) // 15`,
+			},
+			{
+				name: 'avg',
+				description: 'Calculate average of numbers',
+				exportName: 'average',
+				subpath: 'numbers',
+				example: `const mean = average([10, 20, 30]) // 20`,
+			},
+			{
+				name: 'random',
+				description: 'Generate random number',
+				exportName: 'randomInt',
+				subpath: 'random',
+				example: `const num = randomInt(1, 100) // 42`,
+			},
+			{
+				name: 'round',
+				description: 'Round to decimal places',
+				exportName: 'roundTo',
+				subpath: 'numbers',
+				example: `const rounded = roundTo(3.14159, 2) // 3.14`,
+			},
+			{
+				name: 'clamp',
+				description: 'Clamp between min/max',
+				exportName: 'clamp',
+				subpath: 'numbers',
+				example: `const clamped = clamp(150, 0, 100) // 100`,
+			},
+		],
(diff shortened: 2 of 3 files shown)


rtorcato (commented):

Reviewed against the stated risk shape (catalog key reaching a shell/eval/dynamic import/file write/path construction), leaked secrets, and dependency changes.

**Injection / shell / path construction** — No risk. `add`'s `checkbox` prompt constrains selection to the predefined `CliFunction` objects from `functionCategories` (`value: func`, not free text), so nothing user-typed reaches `subpath` or `exportName`. Both are compile-time string literals in the new `src/cli/catalog.ts`. The only place a `subpath` is dynamically imported (`` import(`../${subpath}/index.ts`) `` in `src/cli/catalog.test.ts`) draws from that same hardcoded catalog at test time — not runtime, not user input, and the file isn't shipped (test-only).

**File writes / package.json** — `add` only builds strings and `console.log`s them (import statement + usage examples). No `fs` writes, no `package.json` mutation, so no clobber or traversal surface exists in this command at all.

**Dependencies / build / publish path** — Diff touches only `src/cli/catalog.ts` (new), `src/cli/catalog.test.ts` (new), and `src/cli/cli.ts` (refactor of existing logic to consume the catalog). No `package.json` changes, no new imports outside the existing `commander`/`@inquirer/prompts`/`chalk`/`gradient-string` already in use, and no changes to build scripts. Matches the PR's own claim of no dependency/build changes.

**Secrets** — None. The catalog's `example` strings (`generateSecureToken(32)`, `isStrongPassword(...)`) are illustrative placeholder code, not real credentials.

Correctly resolves #186: `exportName`/`subpath` are now per-entry (handles categories spanning multiple modules), and `catalog.test.ts` asserts every `exportName` is actually exported by `src/<subpath>/index.ts`, closing the drift the issue called out.

No findings at any severity.

### Before merging
Nothing.

rtorcato (commented):

**Scope:** read `gh pr view 200`, `gh pr diff 200`, `gh issue view 186`, and `CLAUDE.md` only — diff-scoped, no repo exploration.

**1. Printed import statements** — Cross-checked all 20 `catalog.ts` entries against the issue's own "printed → real export → module" table: `avg→average/numbers`, `random→randomInt/random`, `round→roundTo/numbers`, `title→titleCase/strings`, `pad→padStart/strings`, `now→nowIso/datetime`, `between→daysBetween/date`, `exists→fileExists/file`, `ext→getFileExtension/file`, `password→isStrongPassword/security`, `token→generateSecureToken/security`, `url→isUrl/validation`, `pid→getProcessId/process`, `uptime→getProcessUptime/process`, `ci→isCI/process`, `node-version→getNodeMajorVersion/node` — every entry matches. The `bySubpath` grouping (`Map<string, string[]>` keyed by `func.subpath`) correctly merges same-subpath selections into one `import { a, b } from '...'` line and emits a separate line per distinct subpath, e.g. `math` spanning `numbers` + `random`. Single-helper selection degenerates to one line correctly. No package-root import is ever printed. This is the actual fix for #186.

**2. `catalog.test.ts`** — Not a shape-only check. It does `await import(\`../${subpath}/index.ts\`)` (a real dynamic import of the actual module) and asserts `Object.keys(mod)).toContain(exportName)`. If any entry named a non-existent export, that module's `it.each` case would genuinely fail — `Object.keys` on the real namespace object won't contain a fabricated name. This is a legitimate regression guard against the exact drift #186 describes.

**3. `generateUsageExample()` deletion** — Confirmed clean: the only caller (inside `add`'s `functions.forEach`) is removed in the same diff and replaced with `console.log(chalk.white(func.example))`, reading the example directly off the catalog entry next to the export name it uses (per the PR's own rationale: "so the two can't disagree"). No usage-example output disappears — every entry in the new catalog carries an `example` field. Nothing else in the diff references the deleted function.

**4. Module-boundary freeze** — Checked all 20 `exportName`/`subpath` pairs for duplicates across entries; none repeat. Each helper has exactly one `(exportName, subpath)` pairing in the catalog, consistent with the one-home-per-helper decision from #184.

**Minor, non-blocking:** `functions[0]?.name` in the final tip line uses optional chaining even though the preceding `if (!functions.length) return` already guarantees a non-empty array — harmless, not worth a round trip.

No defects found. The PR's own verification (`tsc`, `biome check`, `vitest run` 402 passed, and a pty-driven `add` run showing the grouped import output) lines up with what the diff shows.

### Before merging
Nothing.

PR #249, opened 2026-08-28
chore: unpin the Biome version from the $schema URLs

### LINKED ISSUE #248: biome.json pins a Biome version in its $schema URL
## What

```json
"$schema": "https://biomejs.dev/schemas/2.5.7/schema.json"
```

Harmless right now — the installed `@biomejs/biome` **is** 2.5.7, so they agree. The problem is that they only agree by coincidence, and nothing keeps them agreeing.

## Why it matters

Dependabot bumps `@biomejs/biome` on its own schedule. The moment it lands 2.6.x, the config still points at the 2.5.7 schema and every Biome run is validating against the wrong version — which means `pnpm check`, `pnpm lint`, and the husky pre-commit hook, i.e. the commit path for the whole repo.

It fails at the least convenient moment: in a Dependabot PR, where the natural read is "the new Biome broke something" rather than "our schema URL is stale".

## Fix

`repo-tooling doctor` (3.35.0) flags this and offers the fix:

> biome.json pins Biome 2.5.7 in $schema — the next @biomejs/biome bump past it breaks every Biome run
> hint: Run `npx @rtorcato/repo-tooling fix biome` to drop the version from the $schema URL

Run `npx @rtorcato/repo-tooling fix biome --diff` first to confirm the change, then apply. Verify `pnpm check` still passes and that the `extends` of `@rtorcato/repo-tooling/biome` and the local `files.includes` overrides survive untouched.

Found while upgrading repo-tooling in #247, deliberately left out of that PR to keep it reviewable as a plain dependency bump.

rtorcato:
Both `biome.json` files pinned `https://biomejs.dev/schemas/2.5.7/schema.json`, which matched the installed `@biomejs/biome` only by coincidence. The next Dependabot bump past 2.5.7 would have left every Biome run — `pnpm check`, `pnpm lint`, the husky pre-commit hook — validating against a stale schema, and it would have surfaced inside a dependency-bump PR where it reads as "the new Biome broke something".

Swapped the pinned version for `/latest/` in:

- `biome.json`
- `apps/docs/biome.json` (same pin, not mentioned in the issue)

Done by hand rather than through `npx @rtorcato/repo-tooling fix biome`. Its `--diff` showed the scaffold also rewrites the file wholesale, dropping the local `files.includes` exclusions (and, in the docs config, the `overrides` block for `custom.css`). The one-line schema swap is the whole fix.

`pnpm check` (145 files, clean) and `pnpm typecheck` both pass; `extends` and the local overrides are untouched.

Closes #248


rtorcato (commented):

Two-line diff swapping a pinned Biome schema version for `/latest/` in `biome.json` and `apps/docs/biome.json`. The `$schema` field is only consumed by editors/IDEs for JSON-schema validation of the config — it is never fetched or executed at build, lint, or commit time by Biome itself. No code execution path, no secrets, no shell/SQL construction, and no dependency or lockfile changes.

No injection or supply-chain surface here: it's a static URL string in local config, not a command, script, or package reference.

### Before merging
Nothing.

rtorcato (commented):

Minimal, correct diff: both `biome.json` and `apps/docs/biome.json` swap the pinned `2.5.7` schema URL for `/latest/`, exactly as issue #248 requested. Scope matches the PR description (2 files, 2 lines), `extends` and local `files.includes`/`overrides` are untouched, and the PR body documents `pnpm check`/`pnpm typecheck` passing. Skipping `repo-tooling fix biome` in favor of the hand edit is justified — that tool's scaffold would have clobbered the local overrides.

### Before merging
Nothing.

PR #250, opened 2026-08-28
perf(url): validate with URL.canParse instead of try/catch

### LINKED ISSUE #244: isValidUrl can use URL.canParse instead of try/catch
## What

`src/url/index.ts` validates by constructing a URL and catching the throw:

```ts
export function isValidUrl(str: string): boolean {
	try {
		new URL(str)
		return true
	} catch {
		return false
	}
}
```

`URL.canParse(str)` is the native for exactly this. Confirmed available at the `engines.node` floor — probed on Node v22.23.1, `typeof URL.canParse === 'function'` (it landed in 18.17).

## Why bother

Not style. The current form constructs and throws an `Error`, with stack capture, on **every** invalid input — so validating a list of untrusted URLs pays a throw per bad entry. `canParse` returns a boolean without allocating. Also drops a `catch {}` that would silently swallow anything else `new URL` might throw.

## Scope — non-breaking

Internal change only. `isValidUrl` stays exported with the same signature. It is **not** a deprecation candidate: the name reads better at the call site than `URL.canParse`, and `validation.isUrl` delegates to it. This is the `@deprecated`-vs-keep distinction in `skills/feature-audit/SKILL.md` rule 4 — this one keeps its place.

Ships as a `fix:` or `refactor:`, no major.

## Not in scope

Checked and deferred, all above the `>=22` floor (confirmed `undefined` on Node v22.23.1):

- `regex.escapeRegExp` → `RegExp.escape` — Node 24. Also not a drop-in: the native escapes more aggressively, so output strings differ.
- `numbers.sum` → `Math.sumPrecise` — ES2026, Node 24+. Different (exact) float semantics, so a behaviour change rather than a swap.
- `date` / `datetime` / `time` → Temporal — Stage 4 (ES2027), needs Node 26. Revisit when the floor rises; the action then is `@deprecated` tags, not deleting three modules.

Found by the `skills/feature-audit` skill.

rtorcato:
## What

`isValidUrl` now returns `URL.canParse(str)` instead of constructing a `URL` inside a try/catch.

`src/validation/index.ts`'s `isUrl` carried a byte-identical body, so it gets the same swap — fixing only the path the issue names would leave the sibling still throwing per invalid input.

Docs prose in `apps/docs/docs/modules/url.md` described the old implementation; updated.

## Why

The old form constructs and throws an `Error`, with stack capture, on **every** invalid input — validating a list of untrusted URLs paid a throw per bad entry. Also drops a `catch {}` that would have silently swallowed anything else `new URL` might throw.

`URL.canParse` landed in Node 18.17, well under this package's `engines: node >=22` floor.

## Scope

Non-breaking. Both functions keep their signature, export, and behaviour — existing tests in `src/url/index.test.ts` and `src/validation/index.test.ts` pass unchanged.

## Verification

`pnpm typecheck`, `pnpm test` (386 passed), `pnpm check` — all clean.

Closes #244


rtorcato (commented):

`URL.canParse` is spec-defined to return `true` iff `new URL(str)` would succeed — same acceptance set, no behavior drift. Both call sites (`isValidUrl`, `isUrl`) only return a boolean; the result isn't interpolated into a shell command, SQL string, or template, so there's no injection surface here regardless of which parser backs it. No secrets, no new/changed dependencies, no dynamic code execution.

The docs prose was updated to match, and the fix is applied uniformly to both `isValidUrl` and its sibling `isUrl` (no half-fixed duplicate left behind).

### Before merging

Nothing.

rtorcato (commented):

Both call sites (`src/url/index.ts#isValidUrl` and `src/validation/index.ts#isUrl`) were byte-identical try/catch bodies, and both got swapped to `URL.canParse(str)` — the root-cause fix, not just the one the issue named. Signatures and exported behaviour are unchanged: no base URL is passed in either old or new form, so validity semantics match exactly. `URL.canParse` is a static, non-throwing check (landed Node 18.17, well under this repo's `>=22` floor), so dropping the `catch {}` doesn't lose any error handling that mattered — the old catch only ever fed back into a boolean anyway. Docs prose in `apps/docs/docs/modules/url.md` was updated to match. No new abstractions, no test changes needed since behavior is unchanged.

### Before merging

Nothing.

Issue #239, opened 2026-08-20
Four near-duplicate exports escaped the 3.0 boundary audit


rtorcato:
`MODULE-BOUNDARIES.md` has a "Near-duplicates resolved" section for exactly this: *different names for the same function — invisible to a collision check, and the reason an audit was worth doing at all* (cf. `random.randomInt` vs `numbers.getRandomInt`).

Four pairs survived that audit. `scripts/check-readme-exports.mjs` check #2 compares **names**, so it is structurally blind to them.

`holding` because resolving any of these means deleting an export, and the freeze makes that a major version. Not `ai-ready` — the *which one wins* calls are judgement, not mechanics.

## The pairs

| A | B | Difference |
|---|---|---|
| `emails.isValidEmail` (src/emails/index.ts:15) | `validation.isEmail` (src/validation/index.ts:90) | none — same regex, parameter named `email` vs `str` |
| `os.getOsPlatform` (src/os/index.ts:11) | `process.getProcessPlatform` (src/process/index.ts:51) | none — byte-identical body and signature |
| `url.isValidUrl` (src/url/index.ts:13) | `validation.isUrl` (src/validation/index.ts:101) | none — byte-identical body and signature |
| `crypto.randomHex` (src/crypto/index.ts:33) | `security.generateSecureToken` (src/security/index.ts:120) | body identical, **defaults differ**: `length = 16` vs `length = 32` |

Two of these have live maintenance cost today:

- The email pair duplicates a **ReDoS rationale** in a hand-synced comment. Both carry a note explaining that the domain parts exclude `.` so the match stays linear on `'a@!.!.!.!.'` (CodeQL `js/polynomial-redos`). One copy getting "simplified" by someone who didn't read the other reintroduces the finding.
- `crypto.randomHex()` returns 32 hex chars and `security.generateSecureToken()` returns 64 — the bodies are identical but the two are **not** interchangeable at their defaults. Whichever name survives, one set of callers changes token length. That is the one that actually needs care.

The `getProcessPlatform` row is a repeat offence: `MODULE-BOUNDARIES.md` already resolved `getProcessUptime` in favour of `./process` over `./node` for the same reason (guarded `process` access). `getOsPlatform` is the same shape and got missed.

## Two independent pieces of work

**1. Extend the guard — non-breaking, can land any time.**

[...]

PR #252, opened 2026-08-28
refactor: collapse three near-duplicate pairs into delegating aliases

rtorcato:
Resolves three of the four pairs in #239 **without a major**. `Refs #239`, not `Closes` — the fourth pair stays open.

## The decision

#239 framed this as "resolve = delete an export = needs a major". That's true of *deletion*, but deletion isn't the only resolution. A major spent solely on removing aliases charges every consumer a version bump for nothing they asked for.

So: the loser keeps its name and **delegates** to the winner, marked `@deprecated` with the replacement named.

| Winner | Delegating alias | Why this owner |
|---|---|---|
| `emails.isValidEmail` | `validation.isEmail` | `./emails` is the subject-matter home, as `pluralize` went to `./strings` over `./i18n` |
| `url.isValidUrl` | `validation.isUrl` | `./url` owns it, same reasoning |
| `process.getProcessPlatform` | `os.getOsPlatform` | follows the `getProcessUptime` precedent already in MODULE-BOUNDARIES.md |

Owners are the ones #239 itself proposed. Deleting the aliases is still the end state — it just rides along with the next major that happens for a real reason.

## Why now, rather than "leave them, the allowlist documents them"

**Check 4 never protected these pairs.** It fires on *identical* bodies. The moment someone "simplified" one copy, the bodies diverge and the check goes **quiet** — the guard's silence is indistinguishable from success.

That's exactly the risk #239 names for the email pair: both copies carry a hand-synced comment explaining that domain parts exclude `.` so the match stays linear on `'a@!.!.!.!.'` (CodeQL `js/polynomial-redos`). One implementation now, so there's nothing to keep in sync.

## The pair left alone, deliberately

`crypto.randomHex` (default 16 bytes) and `security.generateSecureToken` (default 32) have identical bodies but different defaults — delegating either direction silently halves or doubles the token length callers get. It stays the **sole** entry on `ACCEPTED_DUPLICATES`, with the reason inline. That's a judgement call plus an accepted behaviour change, which is yours, not an agent's.

Allowlist drops 4 → 1.

## Verification

- **Behaviourally identical** — differential-checked `isEmail`/`isValidEmail` and `isUrl`/`isValidUrl` across valid, invalid, ReDoS-shaped and empty inputs, and `getOsPlatform`/`getProcessPlatform`. Zero mismatches.
- **393 tests pass** (53 files), typecheck and biome clean, `build-prod` clean.
- **Costs nothing at consumer bundle size.** The build hoists a delegated body into a shared chunk rather than copying it:

```
validation  577b ["../chunks/chunk-7JI6PUXE.js","../chunks/chunk-BTQDSDRX.js"]
emails      194b ["../chunks/chunk-BTQDSDRX.js"]
url         205b ["../chunks/chunk-7JI6PUXE.js"]
os          798b ["../chunks/chunk-HFYUD75W.js"]
process     201b ["../chunks/chunk-HFYUD75W.js"]
```

Cross-subpath imports already had precedent — `logger` imports `isDev` from `env`.

## Release impact

`refactor:` cuts **no release**. Nothing publishes on merge.


rtorcato (commented):

Pure delegation refactor — no injection, no shell/SQL, no secrets, no dependency changes. `isEmail` now calls the untouched `emails.isValidEmail`, so the ReDoS-safe regex (domain parts exclude `.`, linear on `'a@!.!.!.!.'`) is unchanged — it just has one copy instead of two hand-synced ones, which removes the exact divergence risk #239 flagged rather than introducing one. `getOsPlatform` → `getProcessPlatform` and `isUrl` → `isValidUrl` are likewise delegating to pre-existing, unmodified implementations; PR claims differential testing across valid/invalid/ReDoS-shaped/empty inputs with zero mismatches. `ACCEPTED_DUPLICATES` allowlist shrink is consistent with the collapses made.

### Before merging
Nothing.

rtorcato (commented):

Delegating instead of deleting deviates from #239's "delete outright, needs a major" plan, but it's the repo owner making exactly the judgement call #239 reserved for a human, argued transparently in the PR body. The mechanics hold: `isEmail`/`isUrl`/`getOsPlatform` now forward to the real implementation, each marked `@deprecated` pointing at the replacement, and `ACCEPTED_DUPLICATES` is correctly trimmed since the bodies no longer collide under check 4. `crypto.randomHex` vs `security.generateSecureToken` is rightly left alone given the differing defaults. `Refs #239` (not `Closes`) correctly keeps that pair's issue open.

### Before merging

Nothing.

PR #254, opened 2026-08-28
docs: mark the three deprecated aliases as deprecated

### LINKED ISSUE #253: Docs and README still recommend the three exports #252 deprecates
Follows #252 (which deprecates `validation.isEmail`, `validation.isUrl` and `os.getOsPlatform` into delegating aliases, per #239). The code will carry `@deprecated`; the docs will still recommend all three as the normal way to do the job.

**Blocked on #252 merging** — do not start before then, or a branch off `main` won't have the deprecations to describe.

## The root cause, and why `pnpm docs:generate` alone won't fix it

`scripts/generate-module-docs.mjs` has **no `@deprecated` handling**. Its summary column comes from `firstSentence(jsdoc)`, so re-running the generator reproduces the current rows verbatim — `isEmail` keeps reading "Checks if a string is a valid email address (simple regex)." with nothing marking it deprecated.

So the generator needs to learn the tag first, otherwise every future deprecation silently reads as a recommendation. Suggested: detect `@deprecated` in the JSDoc and prefix the summary (e.g. `**Deprecated.** Use \`isValidEmail\` from \`./emails\`. …`), carrying the replacement through from the tag text rather than hardcoding it.

## Hand-written spots the generator will never touch

| Where | What |
| --- | --- |
| `README.md:192` | `import { getOsPlatform, getOsArch, getHomeDir } from '@rtorcato/js-common/os'` — the System & Process sample teaches the deprecated name |
| `apps/docs/docs/modules/os.md:11,17` | Hero example imports `getOsPlatform` and demonstrates its return values |
| `apps/docs/docs/modules/os.md:25` | Second sample, same |
| 8 × `apps/docs/docs/modules/*.md` | The "see also" line `- [validation](./validation.md) — type guards — \`isString\`, \`isNumber\`, \`isUrl\`` points readers at the deprecated alias. `isUrl` should become a non-deprecated guard, or the line should point at `./url` |

`apps/docs/docs/api/` is typedoc-generated — check whether it renders `@deprecated` natively after #252 lands before changing anything there.

## Scope

Docs only. **Do not** touch `src/`, and do not remove any export — deletion is the still-open half of #239 and needs a major.

## Acceptance

- `pnpm docs:generate` marks the three aliases as deprecated in the generated tables, and re-running it is a no-op afterwards.
- No hand-written sample in `README.md` or `apps/docs/docs/**` teaches a deprecated name as the default.
- `node scripts/check-readme-exports.mjs` still passes — note the checker won't catch any of this, since deprecation isn't removal and all three names still exist.

rtorcato:
Follows #252, which turned `validation.isEmail`, `validation.isUrl` and `os.getOsPlatform` into `@deprecated` delegating aliases. The docs still recommended all three as the normal way to do the job.

## The generator learns the tag

`scripts/generate-module-docs.mjs` had no `@deprecated` handling — the summary column was `firstSentence(jsdoc)`, so re-running `pnpm docs:generate` reproduced the old rows verbatim. It now splits the JSDoc on block tags, and when a `@deprecated` tag is present prefixes the cell with `**Deprecated.** Use <replacement>.` taken from the tag text, not from a list kept in sync by hand. Deprecated names are also filtered out of the generated import example.

The tag body is split on block tags rather than matched with a lazy `[\s\S]*?` — under `/m` that stops at the first line end and truncates a wrapped tag.

## Hand-written spots

- `README.md` System & Process sample — `getOsPlatform` out, `getProcessPlatform` in on the `/process` line.
- `apps/docs/docs/modules/os.md` — hero example switched to `getOsArch`, plus a one-line note pointing at `getProcessPlatform`.
- 8 × `apps/docs/docs/modules/*.md` — the "see also" line now names `isDefined` instead of `isUrl`.

`apps/docs/docs/api/` is typedoc output and is not checked in, so there was nothing to change there.

## Scope

Docs only. No `src/` change, no export removed — deletion is the still-open half of #239 and needs a major.

## Verification

- `pnpm typecheck`, `pnpm check`, `pnpm test` (393 passed) — clean
- `node scripts/check-readme-exports.mjs` — passes
- `pnpm docs:generate` run a second time reports `updated 0` and leaves the tree clean

Closes #253


rtorcato (commented):

Docs-only change plus a parser addition to the build-time doc generator. Checked the checklist:

- **Injection/markdown breakout**: `summarize()` extracts `@deprecated` tag text and interpolates it into a table cell, but both the tag-derived `note` and the `description` go through the existing `firstSentence()`, which already pipes its output through `escapeForMarkdownTable()`. No unescaped path from JSDoc text to the generated markdown.
- **ReDoS**: new regexes (`/^\s*\*\s?/`, `/^@(?=\w)/m`) are simple anchored patterns with no nested/overlapping quantifiers — linear, no catastrophic-backtracking risk. The PR's stated motivation (avoiding a lazy `[\s\S]*?` under `/m`) is a correctness fix, not a new vuln.
- **Secrets**: none introduced.
- **Shell/SQL**: script only reads source files via `fs`; no new `child_process` or string-built queries.
- **Dependencies**: none touched.
- Scope matches the linked issue (#253) and CLAUDE.md — docs/script only, no `src/` or export changes, tests added for the new `summarize`/`buildBlock` behavior.

### Before merging

Nothing.

rtorcato (commented):

`summarize()` splits on `/^@(?=\w)/m`, so a wrapped `@deprecated` tag isn't truncated (covered by the added multi-line test). Traced a bare `@deprecated` (no replacement text) by hand: `note` ends up falsy and is filtered, degrading cleanly — correct, though untested. `buildBlock` excludes deprecated names from the example import. The 8 "see also" swaps to `isDefined` check out: it's a real, non-deprecated `validation` export and reads naturally.

### Before merging
Nothing.

PR #266, opened 2026-09-20
feat(try): add tryCatchSync, a synchronous counterpart to tryCatch

### LINKED ISSUE #264: Add a synchronous counterpart to tryCatch in /try
`/try` exports one entry point:

```ts
tryCatch<T, E = Error>(fn: () => Promise<T>): Promise<Result<T, E>>
```

Async only. That means `Result<T, E>` — the interesting part of the module — can only be used at
async boundaries. Sync code that wants the same shape has to either wrap the call in a promise and
`await` it, or fall back to `try`/`catch` and hand-build the `Result`, which defeats the point.

Requesting a sync sibling returning `Result<T, E>` directly:

```ts
const parsed = attempt(() => JSON.parse(raw))
if (!parsed.ok) return
use(parsed.value)
```

Naming: the bundled Claude skill's worked example already calls this `toResult`, so that name has
some precedent even though it never shipped (see the skill sync issue). `attempt` or `tryCatchSync`
would work equally well — the important thing is that one exists, so `Result` is the module's shape
rather than a detail of its async path.

rtorcato:
`Result<T, E>` was only reachable through the async `tryCatch`. `tryCatchSync(fn)` returns a `Result` directly, so sync code (`JSON.parse`, `new URL`, a throwing validator) gets the same shape without wrapping in a promise.

Named `tryCatchSync` rather than `attempt`/`toResult` so the pairing with `tryCatch` is obvious at the call site.

Docs table regenerated via `pnpm docs:generate`; skill export list updated by hand.

Closes #264


rtorcato (commented):

Pure sync try/catch wrapper mirroring the existing `tryCatch`. No shell, SQL, network, filesystem, or eval usage; no new dependencies; no secrets or user-controlled paths involved. Return shape is identical to the audited async version.

### Before merging
Nothing.

rtorcato (commented):

`tryCatchSync` mirrors `tryCatch`'s exact shape (`Result<T, E>`, same `{ data, error }` fields, same `E = Error` default), just dropping the `Promise` wrapper — correct sync counterpart per #264. Uses `export function` per CLAUDE.md convention. Docs table and skill export list both updated to match. Tests cover runtime success/failure (`JSON.parse`) and type-level shape (`expectTypeOf`), consistent with the existing `tryCatch` test files.

No correctness or convention issues found.

### Before merging
Nothing.

rtorcato:
🤖 *Automated — Pass 1 via ai-workflow.*

**Run `pnpm sync:agents` and push.** Both reviews passed, but the required `lint` check is red and this was not visible to them — reviewers are diff-scoped and never see CI.

Failing step: *Check AGENTS.md is in sync with SKILL.md*

```
AGENTS.md is out of sync with SKILL.md — run `pnpm sync:agents`.
Process completed with exit code 1.
```

The branch edited `skills/js-common/SKILL.md` and `AGENTS.md` by hand; `AGENTS.md` is generated from `SKILL.md`, so it must be regenerated rather than edited. The fix may be the generator or its inputs, not the new `tryCatchSync` code — that part is not in question.

rtorcato (commented):

`tryCatchSync` mirrors `tryCatch`'s try/catch shape exactly, just without the `Promise` wrapper. Type tests (`trycatch.test-d.ts`) confirm `Result<T, E>` is returned directly rather than wrapped in a promise, and runtime tests cover both success (`JSON.parse` valid) and failure (`JSON.parse` invalid → `SyntaxError`) branches. `AGENTS.md`, `SKILL.md`, and the generated docs table are all updated consistently with the new export. Naming (`tryCatchSync`) matches what issue #264 suggested and keeps the pairing with `tryCatch` obvious.

### Before merging
Nothing.

rtorcato (commented):

Pure addition of a synchronous `tryCatch` counterpart — no I/O, no shell/SQL construction, no user-controlled input crossing a trust boundary, no new dependencies, no secrets. The regenerated `AGENTS.md`/`SKILL.md` tables are a single-row diff (`try` row gains `tryCatchSync`) consistent with the doc table in `apps/docs/docs/modules/try.md`, and carry no extra content beyond that.

### Before merging
Nothing.

PR #267, opened 2026-09-20
feat(date): add isWeekday predicate

### LINKED ISSUE #263: Add isWeekend / isWeekday to /date
`/date` has `getDayOfWeek`, `addDays` and `subDays`, so the pieces are there — but the predicate
everyone actually wants is missing, and `getDayOfWeek`'s return value has to be interpreted against
a convention the caller has to remember (0 = Sunday).

Generating a weekday-only price series, I wrote:

```ts
const day = cursor.getUTCDay()
if (day !== 0 && day !== 6) dates.push(new Date(cursor))
```

Requesting:

- `isWeekend(date: string | Date): boolean` and `isWeekday(date)`
- optionally `eachDayBetween(from, to)` / `eachWeekdayBetween(from, to)`, which is the loop that
  predicate almost always ends up inside

Worth deciding explicitly whether these are UTC or local — the rest of `/date` mixes
`getUTCDay`-style and local reads, and for anything date-keyed (like an ISO date string from an API)
UTC is the one that doesn't drift by a day for users west of Greenwich.

rtorcato:
`isWeekend` already shipped in `/date`; `isWeekday` was the missing half. Added as its negation so both keep the same local-time convention as the rest of the module (`getDayOfWeek`, `startOfDay`, `isSameDay` are all local).

Also updated the date docs page and the two module summary tables (AGENTS.md, skills/js-common/SKILL.md).

Skipped from the issue: `eachDayBetween` / `eachWeekdayBetween` (listed as optional there) and any UTC variant — adding a UTC-vs-local split to one predicate while the rest of `/date` stays local would be the inconsistency, not the fix. Worth a separate issue if date-keyed UTC reads are actually wanted module-wide.

Closes #263


rtorcato (commented):

`isWeekday` is a pure, dependency-free predicate (`!isWeekend`) over a local-time day-of-week check. No user input reaches a shell, SQL, filesystem, or eval path; no secrets, no new dependencies, no injection surface. Doc/table edits are text-only.

### Before merging
Nothing.

rtorcato (commented):

`isWeekday` is a straightforward negation of the existing `isWeekend`, keeps the same local-time
convention as the rest of `/date`, and follows the `export function` style required by CLAUDE.md.
JSDoc matches the existing pattern, tests cover a weekday/weekend/edge case, and both summary
tables (AGENTS.md, SKILL.md) plus the docs page were updated consistently. Scope is appropriately
narrow — `eachDayBetween`/`eachWeekdayBetween` and any UTC variant were correctly deferred rather
than bolted on inconsistently.

No bugs or convention violations found.

### Before merging

Nothing.

PR #268, opened 2026-09-20
feat(numbers): add variance, stdDev, median and percentile

### LINKED ISSUE #261: Add variance / stdDev / median / percentile to /numbers
`/numbers` has `sum` and `average` but no dispersion helpers, so every numeric consumer writes the
same three lines. Building a technical-analysis app I hand-rolled standard deviation in two separate
files:

```ts
const sd = Math.sqrt(sum(window.map((v) => (v - mean) ** 2)) / period)
```

Requesting:

- `variance(values, { sample?: boolean })` and `stdDev(values, { sample?: boolean })` — the flag
  matters: a rolling-window indicator divides by `n`, a volatility estimate by `n - 1`, and getting
  it wrong is a silent few-percent error rather than a crash;
- `median(values)`;
- `percentile(values, p)` — with the interpolation method documented, since implementations differ.

All are pure, tiny, and fit the existing module's shape.

rtorcato:
Adds the four dispersion helpers `/numbers` was missing, so consumers stop hand-rolling standard deviation.

- `variance(values, { sample? })` / `stdDev(values, { sample? })` — population (divide by `n`) by default, Bessel's correction (`n - 1`) with `{ sample: true }`.
- `median(values)` — delegates to `percentile(values, 50)`.
- `percentile(values, p)` — `p` is 0–100 (clamped), linear interpolation between closest ranks (R-7, matching Excel `PERCENTILE.INC` / NumPy default). Documented in the JSDoc and the module docs since implementations differ.

Empty inputs return `0`, consistent with the existing `average`. Inputs are not mutated. Docs tables (README, AGENTS.md, SKILL.md, docs site) updated.

typecheck, biome check and the full test suite pass.

Closes #261


rtorcato (commented):

Pure numeric helpers (`variance`, `stdDev`, `median`, `percentile`) plus docs updates. No shell, SQL, network, filesystem, or dependency changes; no secrets. `percentile`'s rank/interpolation bounds are correctly reasoned and covered by tests. No injection or supply-chain surface introduced.

### Before merging
Nothing.

rtorcato (commented):

Verified the math independently (variance/stdDev population vs. sample, and the R-7
percentile interpolation) — matches the JSDoc and test expectations exactly. `median`
correctly delegates to `percentile(…, 50)`. Empty-array and single-value (`sample: true`,
divisor 0) guards return `0` consistently, matching `average`'s existing convention.
Input arrays aren't mutated (`percentile` copies before sorting). Function-declaration
style, JSDoc, and docs (README, AGENTS.md, SKILL.md, docs site) all updated per
convention. No new module, so no `package.json` exports change needed. Tests cover the
documented edge cases.

### Before merging
Nothing.

PR #269, opened 2026-09-20
feat(arrays): add partition, sortBy and zip

### LINKED ISSUE #262: Add groupBy / partition / sortBy to /arrays
`/arrays` covers `unique`, `chunk`, `compact`, `shuffle` — but not the grouping helpers that tend to
be the actual reason people reach for a utility package.

Bucketing daily price bars into ISO weeks, I wrote:

```ts
const byWeek = new Map<string, number>()
for (const candle of candles) {
  const { year, week } = getIsoWeekInfo(new Date(candle.date))
  byWeek.set(`${year}-${String(week).padStart(2, '0')}`, candle.close)
}
```

Requesting, in rough priority order:

- `groupBy(items, keyFn)` → `Record<string, T[]>` (or `Map`, if you'd rather preserve key types)
- `partition(items, predicate)` → `[pass, fail]` — the skill already documents this one as existing,
  which is how I went looking for it
- `sortBy(items, keyFn)` — stable, non-mutating, unlike `Array#sort`
- `zip`, `first`, `last`

Worth noting `partition` appears in the bundled Claude skill's "when to reach for it" list but isn't
exported — see the skill sync issue.

rtorcato:
## What

Adds three helpers to `/arrays`, each one with no native equivalent:

| Export | Signature | Notes |
|---|---|---|
| `partition` | `(arr, (value, index) => boolean) => [pass, fail]` | order preserved in both halves |
| `sortBy` | `(arr, keyFn, direction?) => T[]` | stable, non-mutating; keys read once per element |
| `zip` | `(a, b) => [A, B][]` | truncates to the shorter input |

`sortBy` uses decorate-sort-undecorate (map → sort → map), so `keyFn` runs once
per element rather than once per comparison, and ties fall back to the original
index — stable regardless of engine. Keys may be `number | string | bigint | Date`.

## What this deliberately does not add

The issue also asked for `groupBy`, `first` and `last`. **MODULE-BOUNDARIES.md**
records all three as removed in 4.0 in favour of `Object.groupBy`, `at(0)` and
`at(-1)`, and that decision holds: the package's floor is Node 22, where
`Object.groupBy` is available, and it covers the reporter's own ISO-week example
directly. Re-adding them would reverse a binding decision record to ship wrappers
around the platform.

What was actually broken is the discoverability half of the report: the bundled
Claude skill listed `partition` as available when it was not, which is how the
reporter went looking. `arrays.md` now states positively which native call
replaces each removed helper instead of only noting the removal, and the skill,
`AGENTS.md`, `README.md` and the module tables are all back in sync.

## Tests

Unit (order, stability, descending, `Date` keys, truncation, index argument),
type-level (`expectTypeOf` for all three), and two fast-check properties
(partition halves recombine; `sortBy` is a permutation in non-decreasing order).
`pnpm typecheck`, `pnpm check`, `pnpm test` all clean — 402 passed.

Closes #262


rtorcato (commented):

Pure, dependency-free array helpers (`partition`, `sortBy`, `zip`) plus docs/skill sync. No I/O, no shell/SQL construction, no secrets, no new dependencies, no user-controlled paths. `sortBy`'s comparator only touches caller-supplied keys in memory; no injection surface.

### Before merging
Nothing.

rtorcato (commented):

Reviewed `partition`, `sortBy`, `zip` in `src/arrays/index.ts` plus the doc/skill sync.

- `partition` preserves order, passes `(value, index)` to the predicate as documented, matches tests.
- `sortBy` decorate-sort-undecorate is correct: key computed once per element, index tiebreak makes it genuinely stable, and `<`/`>` on `Date` coerces via `valueOf` so date keys sort chronologically, not lexicographically.
- `zip` truncates to the shorter array as documented; pre-sized result array is fine.
- All three follow the `export function` convention; tabs/single-quote/no-semicolon style matches Biome config.
- Docs (`arrays.md`, `README.md`, `AGENTS.md`, `SKILL.md`) are updated consistently, closing the discoverability gap the issue reported without reversing the 4.0 `groupBy`/`first`/`last` removal decision.

No new module, so no `package.json` `exports` change was needed.

### Before merging
Nothing.

PR #270, opened 2026-09-20
feat(numbers): add signed option to formatPercent

### LINKED ISSUE #265: Add a signed option to formatPercent
`formatPercent(value, fractionDigits?)` doesn't force a sign on positive numbers, so anywhere a
change is displayed next to other changes — finance, metrics dashboards, benchmark diffs — the
column reads wrong: `-0.88%` next to `2.14%`, where the convention is `+2.14%`.

That one detail is the only reason a local `format.ts` survives in my project:

```ts
export const fmtPct = (v: number, dp = 2): string =>
  `${v >= 0 ? '+' : ''}${fmtNumber(v, dp)}%`
```

Requesting an options object:

```ts
formatPercent(2.14, { fractionDigits: 2, signed: true })  // "+2.14%"
formatPercent(0, { signed: true })                        // "+0.00%" or "0.00%"?
```

Two things worth pinning down in the implementation: what zero does (I'd argue unsigned, but it
should be deliberate), and whether the input is a fraction (`0.5` → `50%`) or already a percentage
(`50` → `50%`). The current signature doesn't say, and that ambiguity is its own small papercut —
documenting it would help even without the new option.

rtorcato:
## Summary

Adds a `signed` option to `formatPercent` via a backward-compatible options-object
overload: `formatPercent(value, { fractionDigits, signed })` alongside the existing
`formatPercent(value, fractionDigits?)`, which keeps working unchanged.

## Decisions on the two open questions from the issue

- **What zero does:** zero is never signed, even when `signed: true` —
  `formatPercent(0, { signed: true })` returns `"0%"`, not `"+0.00%"`. A signed zero
  reads as noise in a diff column, and this is asserted in tests.
- **Fraction vs percentage input:** unchanged. `value` is still a fraction
  (`0.25` → `"25%"`), matching every existing caller. This was previously undocumented
  ambiguity; it's now stated explicitly in the JSDoc.

Closes #265

## Test plan
- [x] `pnpm typecheck`
- [x] `pnpm test` (421 passed)
- [x] `pnpm check` (biome, clean)
- [x] New tests cover: options-object `fractionDigits`, `signed: true` on positive/negative/zero/-0, default unsigned behaviour


rtorcato (commented):

Pure logic/formatting change to `formatPercent` (options-object overload for `signed`). No shell/child_process, no SQL, no secrets, no new dependencies, no generated files touched.

### Before merging
Nothing.

rtorcato (commented):

`sign = signed && value > 0` checks the raw value's sign, not the rounded/displayed magnitude. A small positive value that rounds to zero at the given `fractionDigits` still gets a `+`:

```ts
formatPercent(0.00001, { fractionDigits: 2, signed: true }) // "+0.00%"
```

That directly contradicts the PR's own stated invariant ("zero is never signed... reads as noise in a diff column") and isn't covered by the new tests, which only exercise literal `0`/`-0`, not values that round to zero. Decide the sign from the formatted magnitude (e.g. `Number(formatted) !== 0`) rather than the raw `value`.

Backward compatibility and the fraction-vs-percentage semantics are both handled correctly and are exercised by tests; no regression there.

### Before merging
Fix sign logic to key off the rounded/displayed value being zero, not the raw input, and add a test for a value that rounds to zero with `signed: true`.

rtorcato (commented):

No injection, secrets, shell/SQL, or dependency changes — this is pure string formatting plus JSDoc/test additions to `formatPercent`. No generated files (`AGENTS.md`) touched.

### Before merging
Nothing.

rtorcato (commented):

Verified the fix: `signed` now keys off `Number(formatted)` (the rounded, displayed value) rather than the raw `value`. Traced both boundary directions — `0.00001` and `-0.00001` at `fractionDigits: 2` both round to `"0.00"` and are correctly returned unsigned via `formatted.replace('-', '')`, matching the new tests. Exact `0`/`-0` also unsigned. Non-zero positive/negative cases still get `+`/native `-` correctly.

Confirmed no regression for existing `formatPercent(value, fractionDigits?)` callers: when `fractionDigitsOrOptions` is a number or `undefined`, `signed` defaults to `false` and the output reduces to the original `` `${(value*100).toFixed(fractionDigits)}%` `` — behaviorally identical to before this PR.

### Before merging
Nothing.

